PR #12, opened 2022-02-15
Useful list lemmas

wkolowski:
Some lemmas I found useful during the ELMO refactoring, but couldn't find in stdpp.

### DIFF
--- a/theories/VLSM/Lib/ListExtras.v
+++ b/theories/VLSM/Lib/ListExtras.v
@@ -2402,3 +2402,12 @@ Proof.
   rewrite 2!elem_of_list_In.
   apply in_map.
 Qed.
+
+Lemma take_app_inv :
+  forall [A : Type] (n : nat) (l l' : list A) (x : A),
+    take n l = l' ++ [x] -> exists n' : nat, n = S n'.
+Proof.
+  induction n as [| n']; intros l l' x H.
+  - contradict H. rewrite take_0. apply app_cons_not_nil.
+  - exists n'. reflexivity.
+Qed.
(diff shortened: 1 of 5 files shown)


theories/VLSM/Lib/ListExtras.v
```diff
@@ -2434,3 +2434,11 @@ Lemma NoDup_subseteq_length [A : Type]
 Proof.
   apply submseteq_length, NoDup_submseteq; assumption.
 Qed.
+
+Lemma elem_of_list_map
```
bmmoore, line 2438: This is an equivalent claim to `elem_of_list_fmap`, because `fmap` (operator `<$>`) on lists is `map`. If we really need the literal `map` in the statement, the proof of this could at least just be applying `elem_of_list_fmap`.

wkolowski: Oh, that `fmap` thing explains why it's impossible to find this lemma when `Search`ing for `elem_of` and `map`.

wkolowski: @bmmoore In the end, I removed this lemma.

traiansf (approved):

theories/VLSM/Lib/ListExtras.v
```diff
@@ -2394,3 +2393,21 @@ Lemma NoDup_subseteq_length [A : Type]
 Proof.
   apply submseteq_length, NoDup_submseteq; assumption.
 Qed.
+
+Lemma take_app_inv :
+  forall [A : Type] (n : nat) (l l' : list A) (x : A),
+    take n l = l' ++ [x] -> exists n' : nat, n = S n'.
+Proof.
+  induction n as [| n']; intros l l' x H.
```
traiansf, line 2401: are you sure you want to use `H` here?

wkolowski: It's a tiny lemma, so I think the exact naming isn't that important.

PR #13, opened 2022-02-16
port to stdpp 1.6.0 and later

palmskog:
Minor changes to make upgrade to later Coq version possible.


traiansf (approved):

palmskog:
@traiansf so do we want to merge this today? I think it's a good idea since it's more compatible than before ("stdpp 1.6.0 or later")

PR #14, opened 2022-02-17
Make VLSM a record, combine non-VLSMType fields into VLSMMachine

bmmoore:
This changes the definition of VLSM from a nested sigT into
a record, and collapses the two records VLSMSign and VLSMClass
into a single VLSMMachine record because that division was never
useful.
(Separating VLSMType from the rest lets us write definitions like
VLSM_incl that can only be applied to two VLSMs with the same
VLSMType, and it was never helpful to talk about two VLSMs with
the same initial messages and states but different transitions.
The other possible use would be if definitions like
pre_loaded_with_all_messages_vlsm could reuse some of the records,
but VLSMClass dependend on VLSMSign and we never wanted to
change just the transition or valid functions while leaving the
initial states or messages alone).

Unfortunately this somehow changed things so that the extremely
limited form of unification used in the `apply` tactic can no longer
directly apply some lemmas to some goals. For example, lemmas with
conclusions about composite_transition no longer work with the
plain `apply` if the goal was (vtransition (free_composite_vlsm IM)),
even though that expression can reduce to composite_transition.

This can usually be fixed by using `rapply`, which is like
`apply` but based on `refine` and uses a less limited unification
algorithm. I now export it from our Preamble (it's from
Coq.Program.Tactics).
If that fails you can also explicitly supply more arguments.

There is not an `rapply lem in H` form, but the special case of
`apply lem in H;assumption.` can become `revert H;rapply lem`.

### DIFF
--- a/theories/VLSM/Core/ByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces.v
@@ -100,13 +100,6 @@ Definition all_messages_s0 := exist (fun s: @state _ all_messages_type => True) 
 Instance all_messages_state_inh : Inhabited (sig  (fun s: @state _ all_messages_type => True)) :=
   {| inhabitant := all_messages_s0 |}.
 
-Definition all_messages_sig
-    : VLSMSign all_messages_type
-    :=
-    {| initial_state_prop := fun s => True
-     ; initial_message_prop := fun m => False
-    |}.
-
 (**
 
 The [transition] function of the [emit_any_message_vlsm] generates the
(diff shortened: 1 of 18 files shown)


bmmoore:
@palmskog Can you figure out why `apply` was working before but now breaks in places? I think it's generally a matter of `apply` not doing enough to unfold definitions or reducing the lets that extract record fields, but that doesn't explain why we used to be able to apply some of these lemmas.

traiansf (approved):
I think it looks good in general, and it's definitely something we wanted to have. However, some tactics seem to need changing in an awkward way, like needing to use the `@` form of a lemma when applying, or needing to use exact instead of apply. Unfortunately, I don't feel I can contribute any suggestions for improvement.

bmmoore:
I think I will merge it as it is then, but I would still like help from Karl to understand the tactic problems.

PR #16, opened 2022-02-21
minimal changes for compatibility with Coq 8.15

palmskog:
Coq 8.15 did some fixes to the argument naming mechanism. Here, I add some minimal changes **inside proofs** to be compatible with all of Coq 8.13, 8.14, and 8.15. Hopefully, this pays off some technical debt beforehand.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -1,4 +1,5 @@
 -Q theories/VLSM VLSM
+-arg -w -arg -deprecated-instance-without-locality
 
 theories/VLSM/Lib/Preamble.v
 theories/VLSM/Lib/SsrExport.v
--- a/theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
@@ -389,7 +389,7 @@ Proof.
     case_decide as HAv; [| trivial].
     cbn in Hc; destruct Hc as [Hsent | Hseeded].
     + unfold lift_sub_state.
-      rewrite lift_sub_state_to_eq with (Hi0 := HAv).
+      rewrite (lift_sub_state_to_eq _ _ _ _ _ HAv).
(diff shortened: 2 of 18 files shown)


_CoqProject
```diff
@@ -1,4 +1,5 @@
 -Q theories/VLSM VLSM
+-arg -w -arg -deprecated-instance-without-locality
```
traiansf, line 2: This doesn't seem like something we should have in the global options. @palmskog what do you think about it?

palmskog, line 2: So just to explain @traiansf, Coq 8.14 and later adds a large number of warnings to specify things like `Global`/`Local` in many places. This is not supported in Coq 8.13, but I can harmlessly suppress the warning in Coq 8.14 and 8.15.

theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
```diff
@@ -389,7 +389,7 @@ Proof.
     case_decide as HAv; [| trivial].
     cbn in Hc; destruct Hc as [Hsent | Hseeded].
     + unfold lift_sub_state.
-      rewrite lift_sub_state_to_eq with (Hi0 := HAv).
```
traiansf, line 392: Doesn't this work anymore? I kind of liked the `width` syntax for rewrite; A bit unfortunate indeed was that sometimes one would need to append a `0` to the named argument.

wkolowski (approved):
Looks ok. I am quite happy that we aren't updating to Coq 8.15 yet.

theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
traiansf: So they fixed the issue with reusing argument names?

palmskog: I think it's going to be fully fixed only in 8.15.1 (static unique argument names everywhere): https://github.com/coq/coq/pull/15537

PR #18, opened 2022-02-23
Documentation generation boilerplate

palmskog:


resources/config.js
```diff
@@ -0,0 +1,79 @@
...
+    "=>": "⇒",
+    "<>": "≠",
+    "<=": "≤",
+    ">=": "≥",
+    "el": "∈",
+    "nel": "∉",
+    "<<=": "⊆",
+    "|-": "⊢",
+    ">>": "»",
+    "<<": "⊆",
```
wkolowski, line 21: Both `<<=` and `<<` map to `⊆`, is this correct?

resources/config.js
```diff
@@ -0,0 +1,79 @@
...
+    "++": "⧺",
+    "===": "≡",
+    "=/=": "≢",
+    "=~=": "≅",
+    "==>": "⟹",
+    "<==": "⟸",
+    "False": "⊥",
+    "True": "⊤",
+    ":=": "≔",
```
wkolowski, line 30: The symbol on the right is completely illegible.

resources/config.js
palmskog: We don't really use `<<`, so it doesn't really matter, but I changed to avoid the collision.

resources/config.js
palmskog: This is the correct UTF8 character for assignments. If it's illegible, you may want to change your local font.

traiansf (approved):

PR #19, opened 2022-02-23
Jenkinsfile: trigger dependent updater

ehildenb:


bmmoore (approved):
I don't know how this dependency updater thing works, but it looks like it's sending the right hash to the right place.

PR #22, opened 2022-03-07
Remove _composite_state and _composite_label indirection

bmmoore:
This removes the indirection of defining the actual composite state
and label types as _composite_state and _composite_label and making
the nice names composite_state and composite_label definitions that
expand to field projections from the VLSMType composite_type.
This helped typechecking unify (composite_state ?IM) against
some (@state _ ?vtype), without needing to know the expected VLSMType.
The need for this indirection was removed by declaring composite_type
a Canonical Structure, so typechecking will try the guess that vtype
should be some (composite_state _) when otherwise stuck on a
unification problem like that, and also by giving VLSMMachine a
"bidirectionality hint" so typechecking record expressions of
VLSMMachines like composite_vlsm_machine will consider the expected
result type (explicitly given as VLSMMachine composite_type) rather
than needing to discover a vtype by unification at all.


traiansf (approved):

palmskog (commented):
Looks reasonable, but maybe we could document at a high level what the `Canonical` declaration actually means? For example, you could say what happens without it.

PR #23, opened 2022-03-07
Added missing RelDecision instance for msg_dep_happens_before

traiansf:


palmskog (approved):
Looks like a reasonable extension.

theories/VLSM/Core/MessageDependencies.v
```diff
@@ -379,12 +379,19 @@ Global Hint Mode FullMessageDependencies ! - - : typeclass_instances.
...
   (message_dependencies : message -> set message)
   (full_message_dependencies : message -> set message)
   `{FullMessageDependencies _ message_dependencies full_message_dependencies}
   .
 
+Global Instance msg_dep_happens_before_dec : RelDecision (msg_dep_happens_before message_dependencies).
+Proof.
+  intros m1 m2.
+  destruct (decide (m1 ∈ full_message_dependencies m2)); [left|right];
+    rewrite <- full_message_dependencies_happens_before; assumption.
```
palmskog, line 392: This looks so simple that one one might even consider writing the body directly without tactics. But not a requirement for merge, just an observation.

traiansf: hm... I didn't though about that. Usually I tend to make decision instances opaque as I had some cases where having them non-opaque generated trouble. But maybe I can use a definition and declare it opaque after :-)

palmskog: One approach is to use `refine`, see 6d3306e.

wkolowski (commented):
I know it's already merged, but looks ok : )

PR #25, opened 2022-03-08
Full node happens before dependencies are observed and refactorings

traiansf:


theories/VLSM/Core/MessageDependencies.v
```diff
@@ -203,14 +182,46 @@ Lemma msg_dep_full_node_reflects_has_been_observed
...
-    has_been_observed X s m ->
-    has_been_observed X s dm.
+    has_been_observed X s m -> has_been_observed X s dm.
 Proof.
   intros dm m Hdm [Hsent|Hreceived].
   - eapply msg_dep_has_been_sent; eassumption.
   - eapply full_node_has_been_received; eassumption.
 Qed.
 
+(** Under full-node assumptions, [msg_dep_happens_before] reflects the [has_been_observed] predicate.
```
palmskog, line 192:
```suggestion
(** Under full-node assumptions, [msg_dep_happens_before] is equivalent to the [has_been_observed] predicate.
```
"Reflection" has a specific meaning in Coq, so we may want to avoid using it unless that's what we mean.

palmskog (approved):
Looks like reasonable proof refactorings and lemma additions.

wkolowski (commented):
Looks good, I will commit some cosmetic changes.

PR #26, opened 2022-03-17
Validating for free composition from message dependencies

traiansf:
Based on the discussion on Wednesday, I've shown that one can reduce the validator property for the free composition to showing that all of the (direct or indirect) message dependencies of a message can be emitted (using their corresponding dependencies).

@denisadiaconescu and @wkolowski I've marked you as reviewers because (1) this could be added to the VLSM paper; and (2) this could potentially be used to prove validation for the MO components.

Anyone else feel free to review.

### DIFF
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -688,7 +688,7 @@ Lemma [basic_VLSM_incl]
     Qed.
 
     (**
-    If all messages desribed by a predicate <<P>> are valid for the free
+    If all messages described by a predicate <<P>> are valid for the free
     composition pre-loaded with messages described by a predicate <<Q>>, then
     any message which can be emitted by a component pre-loaded with <<P>> can
     also be emitted by the free composition pre-loaded with <<Q>>.
@@ -710,7 +710,7 @@ Lemma [basic_VLSM_incl]
 
     (**
     As a specialization of [valid_preloaded_lifts_can_be_emitted], if all
-    messages desribed by a predicate <<P>> are valid for the free composition,
+    messages described by a predicate <<P>> are valid for the free composition,
     then any message which can be emitted by a component pre-loaded with <<P>>
     can also be emitted by the free composition.
     *)
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -687,6 +687,48 @@ Lemma [basic_VLSM_incl]
       - intro; intros. apply lift_to_composite_state_initial. assumption.
     Qed.
 
+    (** If all messages desribed by a predicate <<P>> are valid for the free
+    composition pre-loaded with messages described by a predicate <<Q>>, then
+    any message which can be emitted by a component pre-loaded with <<P>> can
+    also be emitted by the free composition pre-loaded with <<Q>>.
+    *)
+    Lemma valid_preloaded_lifts_can_be_emitted
+      (P Q : message -> Prop)
+      (HPvalid : forall dm, P dm -> valid_message_prop (pre_loaded_vlsm free_composite_vlsm Q) dm)
+      : forall j m, can_emit (pre_loaded_vlsm (IM j) P) m ->
+        can_emit (pre_loaded_vlsm free_composite_vlsm Q) m.
+    Proof.
+      intros j m Hm.
+      eapply VLSM_incl_can_emit.
+      - apply (pre_loaded_vlsm_incl_relaxed _ (fun m => Q m \/ P m)).
+        intuition.
+      - eapply VLSM_full_projection_can_emit; [|eassumption].
+        apply lift_to_composite_generalized_preloaded_vlsm_full_projection.
+        intuition.
+    Qed.
+
+    (** As a specialization of [valid_preloaded_lifts_can_be_emitted], if all
+    messages desribed by a predicate <<P>> are valid for the free composition,
+    then any message which can be emitted by a component pre-loaded with <<P>>
+    can also be emitted by the free composition.
+    *)
+    Lemma free_valid_preloaded_lifts_can_be_emitted
+      (P : message -> Prop)
+      (Hdeps : forall dm, P dm -> valid_message_prop free_composite_vlsm dm)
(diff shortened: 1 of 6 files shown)


theories/VLSM/Core/Validator.v
```diff
@@ -365,6 +365,94 @@ Definition pre_loaded_with_all_messages_validator_component_proj_incl
...
+  `{FullMessageDependencies message message_dependencies full_message_dependencies}
+  {validator : Type}
+  (A : validator -> index)
+  (sender : message -> option validator)
+  .
+
+(** The property of a message of having a sender and being emitable by the
+component corresponding to its sender pre-loaded with the dependencies of the
+message.
+*)
```
wkolowski, line 385: I somewhat don't like this way of formatting doc comments. Maybe we should keep them properly indented? (I will not commit this suggestion myself).

```suggestion
(** The property of a message of having a sender and being emitable by the
    component corresponding to its sender pre-loaded with the dependencies of the
    message.
*)
```

theories/VLSM/Core/Validator.v
```diff
@@ -365,6 +365,94 @@ Definition pre_loaded_with_all_messages_validator_component_proj_incl
...
+Definition emitable_from_dependencies_prop (m : message) : Prop :=
+  match sender m with
+  | None => False
+  | Some v => can_emit (pre_loaded_vlsm (IM (A v)) (fun dm => dm ∈ message_dependencies m)) m
+  end.
```
wkolowski, line 390: Why not make it `Inductive`?

```suggestion
Inductive emittable_from_dependencies_prop' (m : message) : Prop :=
 | efdp : forall (v : validator) (Hsender : sender m = Some v),
            can_emit (pre_loaded_vlsm (IM (A v)) (fun dm => dm ∈ message_dependencies m)) m ->
              emittable_from_dependencies_prop' m.
```

Also, you made a typo: it's "emittable" with two t's.

wkolowski (commented):
Exceptionally good, (almost) nothing to nitpick about.

theories/VLSM/Core/Validator.v
traiansf: I don't like them indented; seems like a waste of horizontal space. if you want, I can put the beginning of the comment on a new line as we do it in other places, like this:
```
(**
The property of a message of having a sender and being emitable by the
component corresponding to its sender pre-loaded with the dependencies of the
message.
*)
```

theories/VLSM/Core/Validator.v
traiansf: Thanks for the emittable tip :-)

traiansf: > Why not make it `Inductive`?
> 

I guess I'm just not used to viewing it that way... 

I would have rather written it as 
```
Definition emitable_from_dependencies_prop (m : message) : Prop :=
  exists (v : validator) (Hsender : sender m = Some v),
    can_emit (pre_loaded_vlsm (IM (A v)) (fun dm => dm ∈ message_dependencies m)) m.
```
I generally find the inductive definitions like the one you proposed harder to read, and sometimes even harder to trust to be equivalent to the ones I proposed, which are, in some sense more straight-forward for a non-Coq user.

palmskog: The main reason for making a predicate `Inductive` or not is whether you want to use `simpl` and `cbn` to reduce the predicate to its constituents or not. In some settings, splitting up a predicate into parts will clutter the context with useless information - `Inductive` will generally prevent this. However, you no longer get simple proofs about the predicate "for free", you have to apply constructors, possibly manually.

traiansf: Ok I've added an inductive definition and proved them equivalent. Indeed, the inductive definition seems easier to use in the proofs. @wkolowski thanks for the suggestion!

theories/VLSM/Core/Composition.v
```diff
@@ -687,6 +687,50 @@ Lemma [basic_VLSM_incl]
       - intro; intros. apply lift_to_composite_state_initial. assumption.
     Qed.
 
+    (**
+    If all messages desribed by a predicate <<P>> are valid for the free
```
denisadiaconescu, line 691: There is a small typos: described instead of desribed

theories/VLSM/Core/Validator.v
```diff
@@ -365,6 +365,115 @@ Definition pre_loaded_with_all_messages_validator_component_proj_incl
...
+  apply free_valid_from_valid_dependencies with (A v); [assumption | clear v Hemit'].
+  eapply FullMessageDependencies_ind; [eassumption |].
+  intros dm Hdm Hdeps.
+  specialize (Hm dm); spec Hm; [right; assumption |].
+  inversion Hm as [v _ ?]; clear Hm.
+  apply free_valid_from_valid_dependencies with (A v); assumption.
+Qed.
+
+(**
+If a node in a composition satisfied the
```
denisadiaconescu, line 462: Small typo: `satisfies` instead of `satisfied`

theories/VLSM/Core/Validator.v
```diff
@@ -365,6 +365,115 @@ Definition pre_loaded_with_all_messages_validator_component_proj_incl
...
+*)
+Definition valid_all_dependencies_emittable_from_dependencies_prop
+  (i : index) : Prop :=
+    forall l s m, input_valid (pre_loaded_with_all_messages_vlsm (IM i)) l (s, Some m) ->
+      all_dependencies_emittable_from_dependencies_prop m.
+
+(**
+If a message can be emitted by a node preloaded with the message's direct
+dependencies, and if all the dependencies of the message are valid for the
+free composition, the the message itself is valid for the free composition.
```
denisadiaconescu, line 425: Small typo: `then` instead of `the`

theories/VLSM/Core/Validator.v
```diff
@@ -365,6 +365,115 @@ Definition pre_loaded_with_all_messages_validator_component_proj_incl
 
 End component_projection_validator.
 
+Section free_composition_validators.
```
denisadiaconescu, line 368: Maybe add a small description at the beginning of the Section.

denisadiaconescu (approved):
Looks good to me. I spotted just some typos.

PR #28, opened 2022-03-23
Validator for free composition characterization

traiansf:
based on Monday's discussion, I added the proof for the complete characterization of the validating property for free composition in terms of the message and all of its dependencies being emittable by their sender preloaded with just their dependencies.


wkolowski (approved):

denisadiaconescu (approved):
Looks good to me.

PR #30, opened 2022-04-04
A handy lemma which states that `input_valid_transition` is deterministic.

wkolowski:
@denisadiaconescu suggested that I move this lemma here from the new ELMO development.


denisadiaconescu (approved):
Looks good to me.

PR #34, opened 2022-04-19
Refactors finishers - `assumption`

wkolowski:


wkolowski (commented):
I have removed a few spurious `simpl` to `cbn` changes that Brandon wouldn't like.

theories/VLSM/Core/Equivocators/MessageProperties.v
```diff
@@ -53,9 +53,9 @@ Proof.
     + apply equivocator_transition_item_project_inv_messages in Hitem'.
       destruct Hitem' as [_ [_ [_ [_ Ha]]]].
       inversion Hjbs; subst.
-      * left. simpl in Ha. simpl.  rewrite Ha. assumption.
-      * specialize (IHtr _ eq_refl H0). right. assumption.
-    + specialize (IHtr i' eq_refl Hjbs). right. assumption.
+      * by left; cbn in *; rewrite Ha.
```
bmmoore, line 56: This would translate more exactly as `cbn in Ha |- *`, but I suppose it can be guessed that only evaluation in `Ha` and in the goal could be relevant because the rest is just `rewrite Ha`. Also `cbn *` takes negligible time here so performance is fine.

theories/VLSM/Lib/StdppListSet.v
```diff
@@ -10,255 +9,201 @@ Definition set := list A.
...
 
 Lemma set_remove_iff (a b : A) (l : set) :
- NoDup l -> (a ∈ (set_remove b l) <-> a ∈ l /\ a <> b).
+  NoDup l -> a ∈ set_remove b l <-> a ∈ l /\ a <> b.
 Proof.
-split; try split.
-- eapply set_remove_1; eauto.
-- eapply set_remove_2; eauto.
-- destruct 1; apply set_remove_3; auto.
+  repeat split.
```
bmmoore, line 121: Rather than introducing a repeat, maybe `split;[split|].`.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -179,16 +168,10 @@ Lemma filter_length_fn {A} P Q
   length (filter P s) <= length (filter Q s).
 Proof.
   induction s; simpl.
-  - lia.
+  - by lia.
```
bmmoore, line 171: `lia` is already a finisher tactic, and we're not redundantly changing it to `by lia` elsewhere.

theories/VLSM/Core/Composition.v
```diff
@@ -1054,7 +1048,7 @@ Proof.
   - rewrite <- exists_finite.
     split; intros [i Hm]; exists i.
     + by exists (exist _ _ Hm).
-    + destruct Hm as [[im Hinit] Him]. subst. assumption.
+    + by destruct Hm as [[im Hinit] [=->]].
```
bmmoore, line 1051: Please format as `[= ->]` so the lexing is obvious.

theories/VLSM/Core/Equivocators/Composition/Projections.v
```diff
@@ -288,24 +276,23 @@ Proof.
...
   destruct Hproject as [Heqv' [eqv [Heqv Hproject]]].
   exists (equivocator_descriptors_update (zero_descriptor IM) (projT1 (l item)) eqv).
   split.
   { intro i. unfold equivocator_descriptors_update. destruct (decide (i = projT1 (l item))).
-    - subst. rewrite equivocator_descriptors_update_eq. assumption.
-    - rewrite equivocator_descriptors_update_neq by assumption.
-      simpl. by rewrite equivocator_state_project_zero.
+    - by subst; rewrite equivocator_descriptors_update_eq.
+    - rewrite equivocator_descriptors_update_neq by done; cbn.
+  by rewrite equivocator_state_project_zero.
```
bmmoore, line 289: This line lost some indentation.

theories/VLSM/Core/Equivocators/Composition/Projections.v
```diff
@@ -453,20 +435,19 @@ Proof.
...
+  split; [| by repeat split]; clear Hchar.
   intro eqv.
-  destruct (decide (eqv = (projT1 (l item)))).
-  - subst. assumption.
-  - rewrite Heqv'. rewrite Hs.
-    rewrite state_update_neq; [|assumption].
-    unfold proper_descriptor. unfold equivocator_descriptors_update.
-    rewrite equivocator_descriptors_update_neq; [|assumption].
-    apply Hproper.
+  destruct (decide (eqv = (projT1 (l item)))); subst; [done |].
```
bmmoore, line 445: Is there any benefit to doing subst in the second branch, instead of following the original structure and doing it only in the first with `; [subst; done|].`?

bmmoore (changes requested):
Looks pretty good, only a few little problems, and some comments.

theories/VLSM/Core/Equivocators/Common.v
```diff
@@ -615,32 +611,27 @@ Definition equivocator_vlsm
 
 Lemma equivocator_vlsm_initial_message_preservation
   : strong_full_projection_initial_message_preservation X equivocator_vlsm.
-Proof.
-  intro; intros. assumption.
-Qed.
+Proof. by red. Qed.
```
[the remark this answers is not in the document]
wkolowski: Manual says that `red` βιζ-reduces the head constant, but in practice it's mostly used for unfolding the head constant's definition.

I agree that maybe it's better to avoid it, so I'll change this to `by intro.`.

theories/VLSM/Lib/StdppListSet.v
wkolowski: I will change this to `split; [split |]`, but in other places I mostly tended to perform changes in the other direction. @palmskog Opinion?

theories/VLSM/Lib/ListExtras.v
wkolowski: Well, `lia` is good with arithmetic goals, but it doesn't subsume `assumption` and can't even prove `x = x` when `x` is not a `nat`/`Z`, so `by lia` is an improvement in many places.

But yeah, in this particular places `by lia` doesn't make any difference, so I'll change this.

theories/VLSM/Core/Composition.v
wkolowski: Ok, I will add a space to all occurrences of `=->` and `=<-` in the whole repo (this is just ~5 places).

theories/VLSM/Lib/ListExtras.v
bmmoore: No, I mean I thought lia had the property of either succeeding completely or failing that makes it a finisher on it's own, so `lia` is already does exactly the same thing as `lia; done` or `by lia` or `lia; anything`.

theories/VLSM/Core/Equivocators/Composition/Projections.v
bmmoore: I suggest this precisely to keep that signal clear, by only using `; subst` when the previous tactic generated equalities. Here there is only an equality in the positive case of the destruct, and not using `subst` on the side avoids the reader wondering where equalities could be coming from in the negative case.

I certainly agree we shouldn't usually leave equalities floating around for "later" (exceptions include using lemmas that need equalities as arguments), but that doesn't apply here. `;[subst |]` is still immediately adjacent in the control flow.

theories/VLSM/Lib/StdppListSet.v
bmmoore: The problem with `split` in particular is that `split` applies to any single-constructor types, so `repeat split` often goes past splitting on visible logical operators to also applying `constructor` to some types the reader probably hasn't memorized. It's not repetition I'm against, but unpredictability.

theories/VLSM/Lib/ListExtras.v
bmmoore: The stuff about what `lia` doesn't handle is why `by try lia` can be necessary.
BTW, my syntax highlighting (emacs with Proof General) has a different color for finisher tactics (at least a predefined list of them) so I had a reminder.

theories/VLSM/Core/Equivocators/Composition/Projections.v
bmmoore, line 289:
```suggestion
    - by subst; rewrite equivocator_descriptors_update_eq.
    - rewrite equivocator_descriptors_update_neq by done; cbn.
      by rewrite equivocator_state_project_zero.
```

bmmoore (approved):
Looks good now.

Note I committed my own suggested change to put back indentation on that one line. I didn't want to leave it hanging over the weekend.

theories/VLSM/Lib/StdppListSet.v
wkolowski: I agree that this feature of `repeat split` should be avoided and I only use `repeat split` when it splits just the visible connectives.

PR #37, opened 2022-04-29
Top sort refactorings

traiansf:
I've added some corollaries to top-sort to allow a simpler usages in places like `Inspector.v`

### DIFF
--- a/theories/VLSM/Lib/StdppExtras.v
+++ b/theories/VLSM/Lib/StdppExtras.v
@@ -176,6 +176,16 @@ Proof. itauto. Qed.
 Example get_maximal_elements2 : get_maximal_elements Nat.le [1; 4; 2; 4] = [].
 Proof. itauto. Qed.
 
+(**
+Returns all elements <<x>> of a set <<S>> such that <<x>> does not compare less
+than any other element in <<S>> w.r.t to a given precedes relation.
+*)
+Definition get_set_maximal_elements
+  `{HfinSetMessage : FinSet A SetA}
+   (precedes: relation A) `{!RelDecision precedes} (s : SetA)
+   : SetA :=
+    filter (fun a => set_Forall (fun b => ~ precedes a b) s) s.
+
 Lemma filter_ext_elem_of {A} P Q
  `{∀ (x:A), Decision (P x)} `{∀ (x:A), Decision (Q x)} (l:list A) :
  (forall a, a ∈ l -> (P a <-> Q a)) ->
--- a/theories/VLSM/Lib/StdppExtras.v
+++ b/theories/VLSM/Lib/StdppExtras.v
@@ -165,22 +165,22 @@ Qed.
 (* Returns all elements X of l such that X does not compare less
    than any other element w.r.t to the precedes relation *)
 
-Definition get_maximal_elements
+Definition maximal_elements_list
   {A} (precedes: relation A) `{!RelDecision precedes} (l : list A)
   : list A :=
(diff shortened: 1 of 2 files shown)


theories/VLSM/Lib/StdppExtras.v
```diff
@@ -176,6 +176,16 @@ Proof. itauto. Qed.
 Example get_maximal_elements2 : get_maximal_elements Nat.le [1; 4; 2; 4] = [].
 Proof. itauto. Qed.
 
+(**
+Returns all elements <<x>> of a set <<S>> such that <<x>> does not compare less
+than any other element in <<S>> w.r.t to a given precedes relation.
+*)
+Definition get_set_maximal_elements
```
wkolowski, line 183: I don't like this name. `get` and `set` are often used together when the code is about getters and setters, but here `get` is more of a placeholder and `set` means, well, a set. Maybe just call it `maximal_elements`?

theories/VLSM/Lib/TopSort.v
```diff
@@ -768,3 +785,66 @@ Proof.
...
+Section sec_simple_top_sort.
+
+Context
+  `{EqDecision A}
+  (precedes : A -> A -> Prop)
+  `{!RelDecision precedes}
+  `{!StrictOrder precedes}
+  .
+
+Local Lemma Forall_True : forall l : list A, Forall (fun _ => True) l.
```
wkolowski, line 808: Does the `Local` here make any difference?

theories/VLSM/Lib/TopSort.v
```diff
@@ -241,6 +241,23 @@ Qed.
 
 End min_predecessors.
 
+Global Instance precedes_P_transitive `{Transitive A preceeds} (P : A -> Prop) : Transitive (precedes_P preceeds P).
+Proof.
+  intros [x Hx] [y Hy] [z Hz]; unfold precedes_P; cbn.
```
wkolowski, line 248: If it were up to me, I would write

```suggestion
intros [x Hx] [y Hy] [z Hz]; compute.
```

@palmskog What is your opinion on `compute`? Is it considered harmful? Or is it ok to use it for unfolding many definitions at once?

wkolowski (approved):
I have committed some formatting changes.

theories/VLSM/Lib/StdppExtras.v
palmskog: Not a big deal, but the "get" moniker could be dropped, and "set" could be a suffixes, like `maximal_elements_set`. Just "maximal elements" is a bit too generic.

theories/VLSM/Lib/TopSort.v
palmskog: `compute` is essentially just a synonym of `cbv`, which is a general parameterized computation tactic. Is it really needed here? I'd suggest to use the less powerful `red`, if possible. Or maybe `by intros [x Hx] [y Hy] [z Hz].` works?

traiansf: I removed `cbn` but left `unfold preceeds_P`. I think it's ok like this, because we make it clear that transitivity is applied on the underlying `preceeds` relation.

theories/VLSM/Lib/TopSort.v
palmskog: I think this one is so trivial we don't need to move it to `ListExtras`, but @wkolowski can make the final call.

wkolowski: Ok, let it be where it is. If it's needed somewhere else, it can be moved later.

PR #42, opened 2022-05-17
Minimal/maximal definitions

traiansf:
These might already exist in some library we use, but I couldn't find them.

### DIFF
--- a/theories/VLSM/Lib/Preamble.v
+++ b/theories/VLSM/Lib/Preamble.v
@@ -337,18 +337,50 @@ Proof.
 Qed.
 
 (**
-An element <<m>> is minimal statisfying a predicate <<P>> w.r.t. a
-relation <<R>>, if <<P m>> holds and there is no message in relation with <<m>>
-for which <<P>> holds.
+A minimal element of a subset <<S>> (defined by a predicate <<P>>) of some
+preordered set is defined as an element of S that is not greater than any other
+element in S.
 *)
-Definition minimal_wrt `(R : relation A) (P : A -> Prop) (m : A) : Prop :=
-  P m /\ (forall m', P m' -> ~R m' m).
+Definition minimal_among `(R : relation A) (P : A -> Prop) (m : A) : Prop :=
+  P m /\ (forall m', P m' -> R m' m -> R m m').
+
+Example minimal_among_le_0 : minimal_among le (fun _ => True) 0.
+Proof.
+  by split; [| lia].
+Qed.
+
+(** A more concise definition of minimality for strict orders.  *)
+Definition strict_minimal_among `(R : relation A) (P : A -> Prop) (m : A) : Prop :=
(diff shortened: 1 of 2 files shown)


theories/VLSM/Lib/Preamble.v
```diff
@@ -336,6 +336,20 @@ Proof.
   intros. intros a b. symmetry. apply Is_true_iff_eq_true.
 Qed.
 
+(**
+An element <<m>> is minimal statisfying a predicate <<P>> w.r.t. a
+relation <<R>>, if <<P m>> holds and there is no message in relation with <<m>>
+for which <<P>> holds.
+*)
+Definition minimal_wrt `(R : relation A) (P : A -> Prop) (m : A) : Prop :=
+  P m /\ (forall m', P m' -> ~R m' m).
```
wkolowski, line 345: This is a little suspicious, as it implies that `R` is antireflexive on the `x`s that satisfy `P`.

The usual definition of minimal element would be something like `P m /\ forall m', P m -> R m' m -> R m m')`:
- https://ncatlab.org/nlab/show/maximal+element
- https://en.wikipedia.org/wiki/Maximal_and_minimal_elements

theories/VLSM/Lib/Preamble.v
```diff
@@ -336,6 +336,52 @@ Proof.
...
 
+(**
+A minimal element of a subset <<S>> (defined by a predicate <<P>>) of some
+preordered set is defined as an element of S that is not greater than any other
+element in S.
+*)
+Definition minimal_among `(R : relation A) (P : A -> Prop) (m : A) : Prop :=
+  P m /\ (forall m', P m' -> R m' m -> R m m').
+
+Example minimal_among_le_0 : minimal_among le (fun _ => True) 0.
```
palmskog, line 347:
```suggestion
Remark minimal_among_le_0 : minimal_among le (fun _ => True) 0.
```
I think "Remark" fits better here, and we could reserve `Example` for tutorial stuff...

theories/VLSM/Lib/Preamble.v
```diff
@@ -336,6 +336,52 @@ Proof.
...
+Proof.
+  by split; [| lia].
+Qed.
+
+(** A more concise definition of minimality for strict orders.  *)
+Definition strict_minimal_among `(R : relation A) (P : A -> Prop) (m : A) : Prop :=
+  P m /\ (forall m', P m' -> ~ R m' m).
+
+Example strict_minimal_among_lt_0 : minimal_among lt (fun _ => True) 0.
```
palmskog, line 356:
```suggestion
Remark strict_minimal_among_lt_0 : minimal_among lt (fun _ => True) 0.
```

wkolowski (approved):
Looks OK now. I think that having a specialized, more convenient version just for strict orders/relations is a pretty good solution here.

PR #43, opened 2022-05-24
weak full projections from component to free composition

traiansf:
I realized we didn't have these generic results (we only had something similat for lifting from sub-compositions).

Once these are merged, the can be used directly for UMO and RUMO.

### DIFF
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -190,41 +190,39 @@ iff it has the [initial_message_prop]erty in any of the components.
     Definition option_composite_initial_message_prop : option message -> Prop
       := from_option composite_initial_message_prop True.
 
-(**
-We can always "lift" state <<sj>> from component <<j>> to a composite state by
-updating an initial composite state, say [s0], to <<sj>> on component <<j>>.
-*)
     Definition lift_to_composite_label
       (j : index)
       (lj : vlabel (IM j))
       : composite_label
       := existT j lj.
 
-    Definition lift_to_composite_initial_state
-      (j : index)
-      (sj : vstate (IM j))
-      : composite_state
-      := state_update (proj1_sig composite_s0) j sj.
-
-    Definition lift_transition_item_to_composite_initial_state
-      (j : index)
-      : vtransition_item (IM j) -> composite_transition_item :=
-      pre_VLSM_full_projection_trace_item_project (type (IM j)) composite_type
-        (lift_to_composite_label j) (lift_to_composite_initial_state j).
-
-    Definition lift_to_given_composite_state
+(**
+We can always "lift" state <<sj>> from component <<j>> to a composite state by
+updating an given composite state to <<sj>> on component <<j>>.
+*)
+    Definition lift_to_composite_state
       (s : composite_state)
       (j : index)
       (sj : vstate (IM j))
       : composite_state
       := state_update s j sj.
 
-    Definition lift_transition_item_to_given_composite_state
+    Definition lift_to_composite_transition_item
       (s : composite_state)
       (j : index)
       : vtransition_item (IM j) -> composite_transition_item :=
       pre_VLSM_full_projection_trace_item_project (type (IM j)) composite_type
-        (lift_to_composite_label j) (lift_to_given_composite_state s j).
+        (lift_to_composite_label j) (lift_to_composite_state s j).
+
+(**
+A specialized version of [lift_to_composite_state] using the initial composite
+state as the base for lifting.
+*)
+    Definition lift_to_composite_state'
+      := lift_to_composite_state (proj1_sig composite_s0).
+
+    Definition lift_to_composite_transition_item'
+      := lift_to_composite_transition_item (proj1_sig composite_s0).
 
     (**
     Composite versions for [plan_item] and [plan].
@@ -349,14 +347,14 @@ the [composite_valid]ity.
       := (@_apply_plan_last _ composite_type composite_transition start a).
     Definition composite_trace_to_plan := (@_trace_to_plan _ composite_type).
 
-  Lemma lift_to_composite_initial_state_preservation
+  Lemma composite_initial_state_prop_lift
     (j : index)
     (sj : vstate (IM j))
     (Hinitj : vinitial_state_prop (IM j) sj)
-    : composite_initial_state_prop (lift_to_composite_initial_state j sj).
+    : composite_initial_state_prop (lift_to_composite_state' j sj).
   Proof.
     intro i.
-    unfold lift_to_composite_initial_state.
+    unfold lift_to_composite_state'.
     destruct (decide (i = j)); subst.
     - by rewrite state_update_eq.
     - rewrite state_update_neq; cbn; [| done].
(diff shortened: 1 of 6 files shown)


theories/VLSM/Core/Composition.v
```diff
@@ -228,6 +228,12 @@ updating an initial composite state, say [s0], to <<sj>> on component <<j>>.
       : composite_state
       := state_update s j sj.
 
+    Definition lift_to_composite_trace'
```
wkolowski, line 231: Why the `'` in the name? There doesn't seem to be anything named `lift_to_composite_trace` neither in `caper-cbc-proofs` nor in `vlsm`.

theories/VLSM/Core/Composition.v
```diff
@@ -742,13 +748,13 @@ Lemma [basic_VLSM_incl]
...
+      (s : composite_state)
+      (Hs : composite_initial_state_prop s)
       (i : index)
       (si : vstate (IM i))
       (Hsi : vinitial_state_prop (IM i) si)
-      : vinitial_state_prop free_composite_vlsm (state_update s i si).
+      : composite_initial_state_prop (state_update s i si).
     Proof.
       intro j. destruct (decide (j = i)); subst.
       - by rewrite state_update_eq.
```
wkolowski, line 760: I have realized that this is quite a common way of reasoning across the codebase: compare two indices and then `rewrite` either `state_update_eq` or `state_update_neq` depending on the result of the comparison.

Maybe this should be automated a bit?

theories/VLSM/Core/Composition.v
```diff
@@ -757,40 +763,104 @@ Lemma [basic_VLSM_incl]
...
-        assert (Hs' : vinitial_state_prop free_composite_vlsm s')
-          by (subst; apply composite_free_update_initial_state_with_initial; done).
-        by apply initial_state_is_valid.
-      - destruct Ht as [[Hps [Hom Hv]] Ht].
-        unfold transition in Ht. simpl in Ht.
-        destruct Hv as [Hv _]. simpl in Hv.
+      induction Hs using valid_state_prop_ind.
+      - by apply initial_state_is_valid, composite_update_initial_state_with_initial.
+      - destruct Ht as [[Hps [Hom Hv]] Ht]; unfold transition in Ht; cbn in Ht.
+        destruct Hv as [Hv _]; cbn in Hv.
```
wkolowski, line 778:
```suggestion
      - destruct Ht as [[Hps [Hom [Hv _]]] Ht]; cbn in Ht, Hv.
```

You can make it a bit more compact.

theories/VLSM/Core/Composition.v
```diff
@@ -757,40 +763,104 @@ Lemma [basic_VLSM_incl]
...
     Qed.
+
+    Lemma lift_to_composite_valid_preservation :
+      forall (i : index) (cs : composite_state),
+      forall l s om, vvalid (IM i) l (s, om) ->
+        composite_valid (lift_to_composite_label i l)
+          (lift_to_composite_state' cs i s, om).
+    Proof.
+      by intros; unfold lift_to_composite_state'; cbn; rewrite state_update_eq.
+    Qed.
```
wkolowski, line 797:
```suggestion
    Proof. by intros; cbn; rewrite state_update_eq. Qed.
```

No need for `unfold` here. And without `unfold`, you can fir this into one line.

theories/VLSM/Core/Composition.v
```diff
@@ -757,40 +763,104 @@ Lemma [basic_VLSM_incl]
...
+      intros; cbn.
+      unfold lift_to_composite_state'; rewrite state_update_eq.
+      by replace (vtransition _ _ _) with (s', om'); rewrite state_update_twice.
+    Qed.
+
+    Lemma lift_to_composite_initial_message_preservation :
+      forall (i : index),
+          forall m, vinitial_message_prop (IM i) m ->
+          composite_initial_message_prop m.
+    Proof.  by intros i m Hm; exists i, (exist _ _ Hm).  Qed.
```
wkolowski, line 816:
```suggestion
    Proof. by intros i m Hm; exists i, (exist _ _ Hm). Qed.
```

Some whitespace crept in.

theories/VLSM/Core/Composition.v
```diff
@@ -757,40 +763,104 @@ Lemma [basic_VLSM_incl]
...
+          valid_state_prop (pre_loaded_vlsm free_composite_vlsm P) cs ->
+          VLSM_weak_full_projection
+            (pre_loaded_vlsm (IM i) P) (pre_loaded_vlsm free_composite_vlsm P)
+            (lift_to_composite_label i) (lift_to_composite_state' cs i).
+    Proof.
+      intros.
+      apply basic_VLSM_weak_full_projection.
+      - intros l s om (_ & _ & Hv) _ _.
+        by split; [apply lift_to_composite_valid_preservation |].
+      - by intros * ? * [_ Ht]; apply lift_to_composite_transition_preservation.
```
wkolowski, line 829:
```suggestion
      - by inversion 1; apply lift_to_composite_transition_preservation.
```

I don't like this elaborate pattern. I think it would be better to more explicitly state that you just want to invert the first non-dependent hypothesis.

theories/VLSM/Core/Composition.v
```diff
@@ -757,40 +763,104 @@ Lemma [basic_VLSM_incl]
...
+      - by intros s Hs; apply pre_composite_free_update_state_with_initial.
+      - intros _ _ m _ _ [Hm | Hp]; apply initial_message_is_valid; [left | by right].
+        + by eapply lift_to_composite_initial_message_preservation.
```
wkolowski, line 832:
```suggestion
        by eapply lift_to_composite_initial_message_preservation.
```

No need for the bullet.

theories/VLSM/Core/Composition.v
```diff
@@ -757,40 +763,104 @@ Lemma [basic_VLSM_incl]
...
+    Qed.
+
+    Lemma lift_to_preloaded_free_weak_full_projection :
+      forall (i : index) (cs : composite_state),
+          valid_state_prop (pre_loaded_with_all_messages_vlsm free_composite_vlsm) cs ->
+          VLSM_weak_full_projection
+            (pre_loaded_with_all_messages_vlsm (IM i)) (pre_loaded_with_all_messages_vlsm free_composite_vlsm)
+            (lift_to_composite_label i) (lift_to_composite_state' cs i).
+    Proof.
+      constructor; intros.
```
wkolowski, line 856: @palmskog What was the consensus on naming hypotheses? Should we do it only if the names are actually used in the proof?

wkolowski (changes requested):
Looks good, I will commit some tiny improvements. Please address the comment about `'` before I accept.

theories/VLSM/Core/Composition.v
traiansf: I was trying to make it consistent with the name for the state. But given that that name was not much better, I decided to change it.

theories/VLSM/Core/Composition.v
wkolowski: @traiansf I have already started to feel the need for that myself, when juggling some lemmas in the UMO PR. Maybe I will open a PR with such a tactic (and some refactorings) sometime in the future.

wkolowski (changes requested):
I still don't like the naming. We need to strike a balance between being descriptive and short.

Another suggestion: maybe the versions that lift to initial composite state should just get a `'` appended to the end of the name for the general version?

And why aren't the initial versions defined in terms of the general versions? Is it just to avoid lengthy `unfold`ing chains?

wkolowski (approved):

PR #47, opened 2022-06-06
VLSM projections split-up

traiansf:
I've decided to split-up the large VLSM projections module by sub-sections.

Note that I've named the module containing [VLSM_full_projection]s as VLSMEmbeddings as I think that we should eventually rename full_projection to embedding; however, I was not sure whether the best moment to do so would be now.


_CoqProject
```diff
@@ -24,6 +24,11 @@ theories/VLSM/Lib/FinSetExtras.v
 theories/VLSM/Core/VLSM.v
 theories/VLSM/Core/Composition.v
 theories/VLSM/Core/Plans.v
+theories/VLSM/Core/VLSMProjections/VLSMPartialProjection.v
+theories/VLSM/Core/VLSMProjections/VLSMProjection.v
```
palmskog, line 28: If there is "partial" projection, maybe the other file/module should be called "total" projection for clarity, i.e., `VLSMTotalProjection.v`?

palmskog (approved):
This looks like a nice split of content, but to help the reader find their way among the different modules, maybe we use the headers like:
```
(** * <directory/hierarchy>: <file header>
```

PR #48, opened 2022-06-07
Induced composition validators

traiansf:
This changes the existing definitions of `validator` and `projection` to more closely reflect the current understanding:
- projection, is induced projection, being just the component pre-loaded with all messages valid for the composition
- validator has two definitions, one internal to a composition and one external
  - a component has the validator property if the validity predicate ensures that a local transition can be lifted to a valid transition in the composition
  - the induced validator corresponding to a component is the minimally constrained version of the component ensuring that a local transition can be lifted to a valid transition in the composition.


wkolowski (approved):
Looks good. I will commit some changes.

theories/VLSM/Core/Equivocators/Composition/Projections.v
```diff
@@ -2172,7 +2172,7 @@ Proof.
     remember (pre_VLSM_projection_finite_trace_project _ _ _ _ _) as tr.
     replace tr with (equivocators_total_trace_project IM trX); [done |].
     subst. symmetry.
-    apply (equivocators_total_VLSM_projection_trace_project IM (proj1 Hpre_tr)).
+    apply (equivocators_total_VLSM_projection_finite_trace_project IM (proj1 Hpre_tr)).
```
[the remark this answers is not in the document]
traiansf: I don't mind it. I think this is old code.

theories/VLSM/Core/ProjectionTraces.v
```diff
@@ -6,337 +6,72 @@ From VLSM.Core Require Import VLSM Plans Composition VLSMProjections Validator.
...
-its projection (Lemma [component_projection]), as well as between the composite
-VLSM pre-loaded with all messages and the original component VLSM preloaded
-with all messages (Lemma [preloaded_component_projection]).
-
-We then study the extension of these definitions and results to infinite traces.
-
-Finally, we prove some consequences of the [projection_friendly_prop]erty for
-the specific case of projecting a trace to a single component.
+Let us fix an indexed set of VLSMs <<IM>> and their composition <<X>>
+using <<constraint>>.
```
palmskog, line 16: The grammar here feels a bit off, do you mean something like:
> Let us fix an indexed set of VLSMs `<<IM>>` and their composition `<<X>>` while assuming the predicate `<<constraint>>`

theories/VLSM/Core/Validator.v
```diff
@@ -130,6 +130,27 @@ then the induced VLSM is a [VLSM_projection] of the source one.
...
 
+Section sec_projection_induced_validator_pre_definitions.
+
+Context
+  {message : Type}
+  {TX TY : VLSMType message}
+  (label_project : @label _ TX -> option (@label _ TY))
+  (state_project : @state _ TX -> @state _ TY)
+  (label_lift : @label _ TY -> @label _ TX)
+  (state_lift : @state _ TY -> @state _ TX)
```
palmskog, line 141: There are a lot of `@...` here. Maybe it's worth doing some local notations like:
```coq
Local Notation labelTY := (@label _ TY).
Local Notation stateTY := (@state _ TY).
```
And so on.

theories/VLSM/Core/Validator.v
```diff
@@ -674,15 +716,172 @@ Lemma composite_projection_induced_validator_is_projection :
...
+  pre_loaded_with_all_messages_vlsm composite_vlsm_induced_projection_validator.
+
+Lemma preloaded_composite_vlsm_induced_projection_validator_iff
+  (P : message -> Prop)
+  (Hinits : forall m,  vinitial_message_prop (IM i) m -> P m)
+  : VLSM_eq
+      (pre_loaded_vlsm composite_vlsm_induced_projection_validator P)
+      (pre_loaded_vlsm (composite_vlsm_induced_validator IM constraint i) P).
+Proof.
+  apply VLSM_eq_incl_iff; split.
```
palmskog, line 802: Maybe code could be compacted here:
```suggestion
  apply VLSM_eq_incl_iff; split; cbn; apply basic_VLSM_strong_incl.
```
And then remove `cbn; apply basic_VLSM_strong_incl.` in each case.

PR #53, opened 2022-06-17
Definition of the `lastn` function and its properties.

wkolowski:
This is needed to prove the equivalence of definitions of `MO_msg_valid` and `MO_msg_valid'` in the New ELMO.

### DIFF
--- a/theories/VLSM/Lib/ListExtras.v
+++ b/theories/VLSM/Lib/ListExtras.v
@@ -2128,29 +2128,33 @@ Proof.
   by intros A l; unfold lastn; rewrite take_0.
 Qed.
 
+Lemma lastn_ge :
+  forall {A : Type} (n : nat) (l : list A),
+    length l <= n -> lastn n l = l.
+Proof.
+  intros A n l Hge.
+  unfold lastn.
+  rewrite take_ge.
+  - by rewrite rev_involutive.
+  - by rewrite rev_length.
+Qed.
+
+Lemma lastn_app_le :
+  forall {A : Type} (n : nat) (l1 l2 : list A),
+    n <= length l2 -> lastn n (l1 ++ l2) = lastn n l2.
+Proof.
+  intros A n l1 l2 Hlt.
+  unfold lastn.
+  rewrite rev_app_distr, take_app_le; [done |].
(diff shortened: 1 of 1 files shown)


theories/VLSM/Lib/ListExtras.v
```diff
@@ -2108,3 +2108,49 @@ Proof.
...
   by rewrite elem_of_list_In, in_prod_iff, <- !elem_of_list_In.
 Qed.
+
+(** ** The function [lastn] and its properties *)
+
+Definition lastn {A : Type} (n : nat) (l : list A) : list A :=
+  rev (firstn n (rev l)).
+
+Lemma lastn_nil :
+  forall {A : Type} (n : nat),
```
palmskog, line 2118: This is OK for now, but we really should factor out `{A : Type}` into a section variable in the near future. Also, why not keep this on same row as the `forall`, since it's so short?

theories/VLSM/Lib/ListExtras.v
```diff
@@ -2108,3 +2108,49 @@ Proof.
...
+Lemma lastn_0 :
+  forall {A : Type} (l : list A),
+    lastn 0 l = [].
+Proof.
+  by intros A l; unfold lastn; rewrite take_0.
+Qed.
+
+Lemma lastn_cons :
+  forall {A : Type} (n : nat) (h : A) (t : list A),
+    lastn n (h :: t) = if decide (S (length t) <= n) then h :: t else lastn n t.
```
palmskog, line 2133: Isn't it better to split this lemma into one part for `S (length t) <= n` and one part for `~ S (length t) <= n`? This gives more flexibility when to do the case analysis.

palmskog (approved):
This is OK for now, but these lemmas like other lemmas in the should use sections better.

theories/VLSM/Lib/ListExtras.v
wkolowski: I like rewriting lemmas without side conditions, so I'll keep it, but I have added one lemma for the second case (the first is already covered by `lastn_ge`) and refactored the proof of `lastn_cons` to use them.

PR #57, opened 2022-06-24
Equivocation definitions based on Message Dependencies

traiansf:


theories/VLSM/Lib/Preamble.v
```diff
@@ -577,6 +585,11 @@ Proof.
   apply compare_eq_dec.
 Qed.
 
+Inductive Comparable `(R : relation A) : relation A :=
+| comparable_eq : forall a b, a = b -> Comparable R a b
+| comparable_ab : forall a b, R a b -> Comparable R a b
+| comparable_ba : forall a b, R b a -> Comparable R a b.
```
wkolowski, line 591: This looks very suspicious.

In the standard library, there is a type `comparison`, defined as

```coq
Inductive comparison : Set :=  Eq : comparison | Lt : comparison | Gt : comparison.
```

Then, there is a predicate `CompareSpec` defined as

```coq
Inductive CompareSpec (Peq Plt Pgt : Prop) : comparison -> Prop :=
  | CompEq : Peq -> CompareSpec Peq Plt Pgt Eq
  | CompLt : Plt -> CompareSpec Peq Plt Pgt Lt
  | CompGt : Pgt -> CompareSpec Peq Plt Pgt Gt.
```

(There's also `CompareSpecT`, which is a `Type` instead of a `Prop`.)

These two together probably are better than your `Comparable`, as you can use `comparison` to define ordinary, non-dependent functions, and then you can use `CompareSpec` to assign meaning to the results of these functions.

Your `Comparable` is like a blend of `comparison` and `CompareSpec` that only applies to `relation`s.

I think you should try to refactor this using `comparison` and `CompareSpec`.

theories/VLSM/Lib/Preamble.v
```diff
@@ -585,6 +598,27 @@ Definition comparable
...
 
+Lemma Comparable_comparable_iff :
+  forall A (R : relation A) (a b : A),
+    Comparable R a b <-> comparable R a b.
+Proof.
+  split; [by intros []; firstorder|].
+  intros [|[|]].
+  - by constructor 1.
+  - by constructor 2.
+  - by constructor 3.
```
wkolowski, line 609:
```suggestion
  by intros [|[|]]; constructor.
```

theories/VLSM/Lib/Preamble.v
```diff
@@ -585,6 +598,27 @@ Definition comparable
...
+Qed.
+
+Lemma tc_Comparable :
+  forall A (R : relation A) (a b : A),
+    Comparable R a b -> Comparable (tc R) a b.
+Proof.
+  intros *; inversion 1.
+  - by constructor 1.
+  - by constructor 2; constructor.
+  - by constructor 3; constructor.
```
wkolowski, line 619:
```suggestion
  by intros *; inversion 1; repeat constructor.
```

wkolowski (changes requested):
I think you should try to refactor the `Unshelve` and also refactor `Comparable` before this PR can be merged. Besides, I think there are quite a few typos. I will commit my proof-engineering-related suggestions.

theories/VLSM/Lib/Preamble.v
traiansf: You are probably right, but then we should probably get rid of the `comparable` definition next to this.

theories/VLSM/Core/Equivocation.v
```diff
@@ -1583,6 +1583,49 @@ Global Program Instance HasBeenObservedCapability_from_sent_received
...
+  (m : message)
+  (Hsent : has_been_sent X s m) :
+  can_emit (pre_loaded_with_all_messages_vlsm X) m.
+Proof.
+  pose (Heq := pre_loaded_with_all_messages_vlsm_is_pre_loaded_with_True X).
+  rewrite (VLSM_eq_can_emit Heq).
+  eapply sent_can_emit; [by apply (VLSM_eq_valid_state Heq) |].
+  Unshelve.
+  2: by cbn; apply preloaded_HasBeenSentCapability.
+  done.
```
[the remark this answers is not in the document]
traiansf: Hm.. apparently I just needed to move the cbn a little earlier. Without it, it was harder for Coq to figure out the precise instance I was referring to.

theories/VLSM/Lib/Preamble.v
palmskog: I agree with @wkolowski that it would be desirable to purge `comparable`/`Comparable` in favor of `comparison` from Stdlib, preferably everywhere. @traiansf could we switch to `comparison`/`CompareSpec` just in this PR and ignore the other stuff from before for now?

wkolowski (approved):
Looks good.

theories/VLSM/Core/Equivocation.v
```diff
@@ -1919,43 +1919,45 @@ Section Composite.
...
+  Definition composite_has_been_directly_observed
     (s : composite_state IM)
     (m : message)
     : Prop
-    := exists (i : index), has_been_observed (IM i) (s i) m.
+    := exists (i : index), has_been_directly_observed (IM i) (s i) m.
 
-  (** 'composite_has_been_observed' is decidable. *)
-  Lemma composite_has_been_observed_dec : RelDecision composite_has_been_observed.
+  (** 'composite_has_been_directly_observed' is decidable. *)
```
wkolowski, line 1932: I think you want `[]` here instead of single quotes.

theories/VLSM/Core/Equivocation.v
```diff
@@ -1354,68 +1354,68 @@ In protocols like the CBC full node protocol, validators often
...
+      all_traces_have_message_prop vlsm item_sends_or_receives (has_been_directly_observed vlsm) s m.
 Proof.
   intros.
   apply prove_all_have_message_from_stepwise; [| done].
-  apply has_been_observed_stepwise_props.
+  apply has_been_directly_observed_stepwise_props.
 Qed.
 
-Lemma proper_not_observed `(vlsm : VLSM message) `{HasBeenObservedCapability message vlsm}:
+Lemma proper_not_directly_observed `(vlsm : VLSM message) `{HasBeenDirectlyObservedCapability message vlsm}:
```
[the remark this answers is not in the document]
traiansf: It might, but I don't think it's relevant here. The point of the `proper_not_directly_observed` result is to say that if a message is not directly observed, then it won't appear as sent or receive on any valid trace leading to s

theories/VLSM/Core/Equivocation.v
```diff
@@ -1919,43 +1919,45 @@ Section Composite.
   End composite_has_been_received.
 
 
-  (** A message 'has_been_observed' for a composite state if it 'has_been_observed' for any of
-  its components.*)
-  Definition composite_has_been_observed
+  (**
+  A message [has_been_directly_observed] for a composite state if it
+  [has_been_directly_observed] for any of its components.
```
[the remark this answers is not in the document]
traiansf: To compromise, I'll change 'for' to 'in'
'by' would imply that the state is the observer, which doesn't feel adequate to me.

theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
```diff
@@ -633,7 +633,7 @@ Proof.
     destruct Hc as [_ [_ [Hc _]]].
     destruct Hc as [Hsent | Hseeded].
     + left.
-      apply composite_has_been_observed_sent_received_iff.
+      apply composite_has_been_directly_observed_sent_received_iff.
```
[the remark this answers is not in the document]
traiansf: The definition composite_has_been_directly_observed s m is given as exists i, has_been_directly_observed (s i) m
This lemma says the definition is equivavalent to the disjunction between composite_has_been_sent and composite_has_been_received

PR #59, opened 2022-07-04
More coqdoc structure, fix a brittle proof

palmskog:
Here I add more coqdoc structure to projections-related files, to make the whole package/project more surveyable.

A proof in `LimitedByzantineTraces` was brittle and broke again on 8.16. I included a permanent fix that makes the proof stable.

### DIFF
--- a/theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
@@ -375,10 +375,8 @@ Proof.
     destruct l as [sub_i li]; destruct_dec_sig sub_i i Hi Heqsub_i; subst sub_i
     ; destruct IHHbyzantine as [[Htr0_ann Hsi_ann] Htr0_eqv_byzantine]
     ; cbn in Htr0_eqv_byzantine |- *.
-    unfold msg_dep_annotate_trace_with_equivocators,
-     coeqv_annotate_trace_with_equivocators, annotate_trace in Htr0_eqv_byzantine.
     remember (@finite_trace_last _ (annotated_type (free_composite_vlsm IM) (set index)) _ _)
-          as lst.
+     as lst in Htr0_eqv_byzantine at 1 |- * at 1 2 3 4 5 6.
     assert (Hlsti : original_state lst = lift_sub_state IM (set_diff (enum index) byzantine)
                                           (finite_trace_last si tr0)).
     {
--- a/theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
@@ -7,7 +7,7 @@ From VLSM.Core Require Import Validator Equivocation Equivocation.FixedSetEquivo
 From VLSM.Core Require Import Equivocation.LimitedEquivocation Equivocation.LimitedEquivocation.
 From VLSM.Core Require Import Equivocation.MsgDepLimitedEquivocation Equivocation.TraceWiseEquivocation.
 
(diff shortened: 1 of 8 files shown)


theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
```diff
@@ -375,10 +375,8 @@ Proof.
     destruct l as [sub_i li]; destruct_dec_sig sub_i i Hi Heqsub_i; subst sub_i
     ; destruct IHHbyzantine as [[Htr0_ann Hsi_ann] Htr0_eqv_byzantine]
     ; cbn in Htr0_eqv_byzantine |- *.
-    unfold msg_dep_annotate_trace_with_equivocators,
-     coeqv_annotate_trace_with_equivocators, annotate_trace in Htr0_eqv_byzantine.
     remember (@finite_trace_last _ (annotated_type (free_composite_vlsm IM) (set index)) _ _)
-          as lst.
+     as lst in Htr0_eqv_byzantine at 1 |- * at 1 2 3 4 5 6.
```
wkolowski, line 379: I never suspected this kind of thing is possible.

theories/VLSM/Core/VLSMProjections/VLSMEmbedding.v
```diff
@@ -6,16 +6,16 @@ From VLSM.Core Require Import VLSM VLSMProjections.VLSMTotalProjection.
...
 
-(** * VLSM Projections : VLSM Full Projection (Embedding)
+(** * VLSM Full Projection (Embedding)
 
 A VLSM projection guaranteeing the existence of projection for all labels and
 states, and the full correspondence between [transition_item]s.
-We say that VLSM X fully projects (embeds) into VLSM Y (sharing the same messages)
-if there exist maps <<label_project>> taking X-labels to Y-labels
-and <<state_project>> taking X-states to Y-states, such that the
+We say that VLSM <<X>> fully projects (embeds) into VLSM <<Y>> (sharing the same messages)
```
wkolowski, line 13: To clarify, we use `<<>>` for all local variables (even "pseudovariables" that don't actually appear in the code), and not only for variables declared using `Variables` or `Context`, right?

theories/VLSM/Core/VLSMProjections/VLSMTotalProjection.v
```diff
@@ -698,7 +699,7 @@ Proof.
   - by apply VLSM_projection_initial_state.
 Qed.
 
-(** ** Projection Friendliness
+(** ** Projection friendliness
```
wkolowski, line 702: Maybe this should be written as "Projection-friendliness"?

wkolowski (approved):

theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
palmskog: For posterity, I'll break down what happens:
- in 8.15, the bare `remember ... as lst` maximally unifies on all occurrences it can see, which turns out to be `remember ... as lst in Htr0_eqv_byzantine at 1 |- * at 1 2 3 4 5 6`.
- in 8.16, the default was changed and `remember ... as lst` no longer does "maximal" unification, but seemingly uses the first occurrence it sees. But if we are explicit with `in ...`, we get the old behavior.

theories/VLSM/Core/VLSMProjections/VLSMEmbedding.v
palmskog: We should use some form of style for these kinds of meta variables, not necessarily `<<>>`. But I mark them up here to make later replacement easier. I'm not sure we need a complete convention for all "variables" yet.

PR #60, opened 2022-07-06
    MessageDependecies necessity refinement; ObservedBeforeSent relation

traiansf:


theories/VLSM/Core/MessageDependencies.v
```diff
@@ -110,13 +111,20 @@ Proof.
     contradict Hinit; apply no_initial_messages_in_X.
   - apply (directly_observed_valid (pre_loaded_vlsm X P) s).
     + exists (Some m). by apply can_produce_valid.
-    + cut (has_been_directly_observed X s dm).
+    + destruct Hproduce as ((pre_s, im) & l & Ht).
+    cut (has_been_directly_observed X pre_s dm).
```
wkolowski, line 115: A more backward proof would be better than using a `cut`. I will take care of it :)

theories/VLSM/Core/MessageDependencies.v
```diff
@@ -245,6 +264,35 @@ Inductive HasBeenObserved (s : vstate X) (m : message) : Prop :=
       msg_dep_happens_before message_dependencies m m' ->
       HasBeenObserved s m.
 
+(**
+A relation capturing the messages <<m1>> directly observed at the moment of
+emitting a message <<m2>>.
+
+Note that this might be different from the [msg_dep_rel]ation: on one hand,
```
wkolowski, line 271: I believe it's "on the one hand". I will fix it.

theories/VLSM/Core/MessageDependencies.v
```diff
@@ -245,6 +264,35 @@ Inductive HasBeenObserved (s : vstate X) (m : message) : Prop :=
...
+on the other hand, it might be that a message can be emitted from a state containing
+more than its required depdendencies.
+*)
+Inductive DirectlyObservedBeforeSend (m1 m2 : message) : Prop :=
+| directly_observed_before_send :
+    forall (s : vstate X) (item : vtransition_item X),
+      input_valid_transition_item (pre_loaded_with_all_messages_vlsm X) s item ->
+      has_been_directly_observed X s m1 ->
+      output item = Some m2 ->
+      DirectlyObservedBeforeSend m1 m2.
```
wkolowski, line 282: Why not a record?

theories/VLSM/Core/MessageDependencies.v
```diff
@@ -391,18 +439,12 @@ Global Instance composite_message_dependencies
...
-    exists (is i, iom), li.
-    revert Ht.
-    apply
-      (VLSM_projection_input_valid_transition (preloaded_component_projection IM _))
-      with (lY := li).
-    unfold composite_project_label; cbn.
-    case_decide as Heqi; [| done].
-    by replace Heqi with (@eq_refl index i) by (apply Eqdep_dec.UIP_dec; done).
+    eapply message_dependencies_are_necessary;
+      [typeclasses eauto | exact Ht | done].
```
wkolowski, line 447: I think we should avoid `exact`, but I checked and `done` doesn't work here.

theories/VLSM/Core/MessageDependencies.v
```diff
@@ -527,6 +569,94 @@ Proof.
...
+definition is that RHS can be emitted by any of the machines in the composition.
+*)
+Inductive CompositeDirectlyObservedBeforeSend (m1 m2 : message) : Prop :=
+| composite_directly_observed_before_send :
+    forall (s : composite_state IM) (item : composite_transition_item IM)
+      (i := projT1 (l item)),
+      input_valid_transition_item RFree s item ->
+      has_been_directly_observed (IM i) (s i) m1 ->
+      output item = Some m2 ->
+      CompositeDirectlyObservedBeforeSend m1 m2.
```
wkolowski, line 583: Again, why not a `Record`?

theories/VLSM/Core/MessageDependencies.v
```diff
@@ -527,6 +569,94 @@ Proof.
...
+Lemma CompositeDirectlyObservedBeforeSend_subsumes_msg_dep_rel
+  `{forall i, MessageDependencies message_dependencies (IM i)} :
+  forall m, can_emit RFree m ->
+  forall dm, msg_dep_rel message_dependencies dm m ->
+    CompositeDirectlyObservedBeforeSend dm m.
+Proof.
+  intros m Hm dm Hdm; apply can_emit_composite_project in Hm as [j Hjm].
+  by eapply composite_DirectlyObservedBeforeSend_lift,
+    DirectlyObservedBeforeSend_subsumes_msg_dep_rel;
+    cycle 1.
```
wkolowski, line 621: This `cycle 1` is not needed. I will remove it.

theories/VLSM/Core/MessageDependencies.v
palmskog: Under the hood, there's not much of a difference from `Inductive` with a Coq record here (they both generate an inductive type, `MutInd` at the term level). For a record to make sense, I think there needs to be something specific you want to project out and name. If you only need to project out one thing, then it may be better to write a standalone function that does a match.

As discussed earlier, a good reason for using just a plain `Inductive` is to prevent automatic unfolding of the predicate, and if you don't need to project out anything, I think avoiding `Record` is fine.

traiansf: In the end, I decided to use records, and here is the reasoning for that:
- I find it easier to understand conjunctions as fields in a record than as implications in an `Inductive`
- There were already records used for similar reasons in the module
- using records parameterized in more arguments allowed for finer-grained results  which could prove useful in other contexts

PR #62, opened 2022-07-15
Sanitation (export, global, local)

DafinaTrufas:


palmskog (approved):
Looks good. @DafinaTrufas when you apply elsewhere, maybe you can add something about this to the coding conventions document as well? Something like:

> Use `#[local]` instead of `Local`, ...

PR #64, opened 2022-07-18
Spell checking

DafinaTrufas:


theories/VLSM/Core/Equivocators/Projections.v
```diff
@@ -250,7 +250,7 @@ descriptor of a projection operation in terms of the input descriptor
 are Existing, then the output index must be less than the input), while also
 guaranteeing that the output state of such a projection has a size less than
 the index of the input descriptor in case that output descriptor becomes
-NewMachine (signaling that the projection is complete).
+NewMachine (signalling that the projection is complete).
```
wkolowski, line 253: Google says "signaling" (one 'l') is the more common variant.

wkolowski (approved):
Looks good. What tool are you using for it?

theories/VLSM/Lib/StreamFilters.v
```diff
@@ -477,7 +477,7 @@ Qed.
 
 (** [stream_filter_positions] produces a [filtering_sequence].
 *)
-Lemma stream_filter_positions_fitering_subsequence
+Lemma filtering_subsequence_stream_filter_map
```
palmskog, line 480: The current suggestion here actually changes the meaning of the lemma name:
```suggestion
Lemma stream_filter_positions_filtering_subsequence
```
Maybe we can do lemma names in a different pull request? They would need more effort to review.

theories/VLSM/Lib/Traces.v
```diff
@@ -16,7 +16,7 @@ Context {A B : Type}.
 
 (** This definition is similar to that for lazy lists from Chapter 13
 of the #<a href="https://github.com/coq-community/coq-art">Coq'Art book</a>#.
-However, to support traces following labeled transition relations, constructors
+However, to support traces following labelled transition relations, constructors
```
palmskog, line 19: I believe "labeled" (one l) is typical US English. Is it perhaps a UK English suggestion?
```suggestion
However, to support traces following labeled transition relations, constructors
```

palmskog (commented):
This is much better now, but perhaps we can save the lemma name changes for another pull request?

theories/VLSM/Lib/StreamFilters.v
traiansf: I'm to blame for that. I suggested to Dafina to fix the spelling for some definitions for which she had fixed it in the CoqDoc references, but it must have been a search/replace error in this particular case.

palmskog (approved):
Now everything looks good, thanks!

PR #66, opened 2022-07-19
Refactored observed_before_send

traiansf:
- Refactored `observed_before_send` to allow transitions receiving and sending at the same time
- Reverted `message_dependencies_are_necessary` to earlier form
- Added `  `{!Irreflexive (msg_dep_happens_before message_dependencies)}` constraint to `MessageDependencies` class

### DIFF
--- a/theories/VLSM/Core/MessageDependencies.v
+++ b/theories/VLSM/Core/MessageDependencies.v
@@ -7,11 +7,18 @@ From VLSM.Core Require Import SubProjectionTraces Equivocation EquivocationProje
 
 (** * VLSM Message Dependencies
 
-An abstract framework for the full-node condition.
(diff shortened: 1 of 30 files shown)


theories/VLSM/Core/MessageDependencies.v
```diff
@@ -677,7 +710,7 @@ Proof.
     [eapply input_valid_transition_preloaded_project_active |..].
 Qed.
 
-Lemma composite_observed_before_send_iff m1 m2 : 
+Lemma composite_observed_before_send_iff m1 m2 :
```
wkolowski, line 713: I rewrote the proof here to have separate bullets for each subgoal, as I think this kind of symmetry is helpful for proofs of equivalence.

PR #67, opened 2022-07-20
Comment-spec-ratio

DafinaTrufas:


palmskog (changes requested):
This looks good, but I have two requests for changes:

- We shouldn't run the scripts *every time* we build the project. It should be an optional task (not part of `all`) that only lives in `Makefile.coq.local`, similar to `coqdoc` - maybe call it `ratios`.

- The file `COMMENT-RATIO.md` should not be checked in to the repo. We can generate it when we need it, and discuss the "latest version" from time to time. In the future, we might put a version in our documentation source repo.

traiansf:
> * The file `COMMENT-RATIO.md` should not be checked in to the repo. We can generate it when we need it, and discuss the "latest version" from time to time. In the future, we might put a version in our documentation source repo.

Then why did we bother to request generating a `.md` file with a table?

palmskog:
> > * The file `COMMENT-RATIO.md` should not be checked in to the repo. We can generate it when we need it, and discuss the "latest version" from time to time. In the future, we might put a version in our documentation source repo.
> 
> Then why did we bother to request generating a `.md` file with a table?

The table and its data is really interesting and I already learned a lot from them, but it's a snapshot and I really don't think we want to update it for *every* commit into the repo. And if we don't update for every commit, then the table will quickly be out-of-date and possibly misleading. 

I see this in the same way as the axiom diagnostics or `make validate` - it's something we generate from time to time and use to make maintenance decisions.

Makefile.coq.local
```diff
@@ -26,6 +26,10 @@ coqdoc: $(DOCGLOBFILES) $(DOCVFILES) $(CSSFILES) $(JSFILES) $(HTMLFILES)
 	$(HIDE)cp $(CSSFILES) $(JSFILES) $(COQDOCDIR)
 .PHONY: coqdoc
 
+ratios:
+	@scripts/all-comment-ratio.sh > COMMENT-RATIO.md
+.PHONY: ratios
```
palmskog, line 31: One final suggestion that is a bit more Makefile-idiomatic:
```suggestion
RATIOS.md: $(DOCVFILES)
	@scripts/all-comment-ratio.sh > RATIOS.md
```
This make it possible to regenerate with `make RATIOS.md` once files are edited.

palmskog (approved):
Thanks, only one final suggestion.

PR #68, opened 2022-07-20
Some lemmas about lastn

wkolowski:
These are needed to prove an alternative version of lemma 3.1 for ELMO, requested by Denisa.

### DIFF
--- a/theories/VLSM/Lib/ListExtras.v
+++ b/theories/VLSM/Lib/ListExtras.v
@@ -2096,7 +2096,7 @@ Proof.
   by rewrite rev_length; lia.
 Qed.
 
-(** [astn] either skips the head of the list or not, depending on how big a suffix we want. *)
+(** [lastn] either skips the head of the list or not, depending on how big a suffix we want. *)
 Lemma lastn_cons :
   forall {A : Type} (n : nat) (h : A) (t : list A),
     lastn n (h :: t) = if decide (S (length t) <= n) then h :: t else lastn n t.
@@ -2139,18 +2139,6 @@ Proof.
   intros A n l.
   unfold lastn.
   by rewrite rev_length, take_length, rev_length.
-Qed.
-
-(** If the <<n>>-th element of <<l>> is <<x>>, then we can decompose long enough
-    suffixes of <<l>> into <<x>> and a suffix shorter by 1. *)
-Lemma lastn_decomposition :
-  forall {A : Type} (n : nat) (l : list A) (x : A),
-    l !! n = Some x -> lastn (length l - n) l = x :: lastn (length l - S n) l.
-Proof.
-  intros A n l x H.
-  unfold lastn.
-  rewrite <- rev_length, <- !skipn_rev, rev_involutive.
-  by apply drop_S.
 Qed.
 
 Program Definition not_null_element
--- a/theories/VLSM/Lib/ListExtras.v
+++ b/theories/VLSM/Lib/ListExtras.v
@@ -2053,9 +2053,11 @@ Qed.
 
 (** ** The function [lastn] and its properties *)
 
+(** [lastn] returns a suffix of length <<n>> from the list <<l>>. *)
 Definition lastn {A : Type} (n : nat) (l : list A) : list A :=
   rev (firstn n (rev l)).
 
+(** If the list is [[]], then the result of [lastn] is also [[]]. *)
 Lemma lastn_nil :
   forall {A : Type} (n : nat),
     lastn n (@nil A) = [].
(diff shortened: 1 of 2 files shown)


theories/VLSM/Lib/ListExtras.v
```diff
@@ -2053,23 +2053,27 @@ Qed.
 
 (** ** The function [lastn] and its properties *)
 
+(** [lastn] returns a suffix of length <<n>> from the list <<l>>. *)
```
palmskog, line 2056: Maybe skip the period for one-liners?
```suggestion
(** [lastn] returns a suffix of length <<n>> from the list <<l>> *)
```

theories/VLSM/Lib/ListExtras.v
```diff
@@ -2081,16 +2085,18 @@ Proof.
...
     n <= length l2 -> lastn n (l1 ++ l2) = lastn n l2.
 Proof.
   intros A n l1 l2 Hlt.
   unfold lastn.
   rewrite rev_app_distr, take_app_le; [done |].
-  rewrite rev_length; lia.
+  by rewrite rev_length; lia.
 Qed.
 
+(** [astn] either skips the head of the list or not, depending on how big a suffix we want. *)
```
palmskog, line 2099:
```suggestion
(** [lastn] either skips the head of the list or not, depending on how big a suffix we want. *)
```

theories/VLSM/Lib/ListExtras.v
```diff
@@ -2101,6 +2107,52 @@ Proof.
   - by rewrite (lastn_app_le _ [h] t); [| lia].
 Qed.
 
+(** If <<l1>> is a prefix of <<l2>>, then the reverse of <<l1>> is a suffix of <<l2>>. *)
+Lemma suffix_rev :
```
palmskog, line 2111: Name can be more informative here.
```suggestion
Lemma prefix_suffix_rev :
```

theories/VLSM/Lib/ListExtras.v
```diff
@@ -2101,6 +2107,52 @@ Proof.
...
+  intros A l n1 n2 Hle.
+  unfold lastn.
+  apply suffix_rev.
+  exists (take (n2 - n1) (drop n1 (rev l))).
+  rewrite take_take_drop.
+  by f_equal; lia.
+Qed.
+
+(** The length of <<lastn n l>> is the smaller of <<n>> and the length of <<l>>. *)
+Lemma length_lastn :
```
palmskog, line 2135: More informative to have:
```suggestion
Lemma length_min_lastn :
```

theories/VLSM/Lib/ListExtras.v
```diff
@@ -2101,6 +2107,52 @@ Proof.
...
+    length (lastn n l) = min n (length l).
+Proof.
+  intros A n l.
+  unfold lastn.
+  by rewrite rev_length, take_length, rev_length.
+Qed.
+
+(** If the <<n>>-th element of <<l>> is <<x>>, then we can decompose long enough
+    suffixes of <<l>> into <<x>> and a suffix shorter by 1. *)
+Lemma lastn_decomposition :
```
palmskog, line 2146: Maybe better to put in `StdppExtras.v`, since uses the Stdpp `l !! n` notation. In a separate future PR, we can probably ensure better separation between `ListExtras.v` and `StdppExtras.v`. Also, the name is not that informative, maybe something like `lastn_length_cons`?

palmskog (approved):
Looks reasonable, but I have some suggestions.

theories/VLSM/Lib/ListExtras.v
wkolowski: I think I will keep it. After all, this would need to be decided based on how it's displayed, not on how long it is in the code (all these `[]` and `<<>>` could make a difference).

theories/VLSM/Lib/ListExtras.v
wkolowski: I think `length_lastn` is better for mnemonic reasons. When I name things, I try to mimic their syntax, so a lemma which tells me what is the `length` of `lastn n l` would be called `length_lastn`. Adding `min` in there messes the mnemonic up, not to mention it means you need to remember what the actual result is...

This also applies to the previous comment. When my lemma has conclusion `suffix (rev l1) (rev l2)`, I would name it `suffix_rev`, so that I can reconstruct the name just by looking at the goal I need to prove. Adding `prefix` there just makes the mnemonic less robust, because I need to remember that the hypothesis is actually `prefix` and not something else.

theories/VLSM/Lib/ListExtras.v
wkolowski: Ok, I will move it there.

Regarding the name, this is the kind of lemma that I do not have any mnemonic system for, as it's not "syntax-driven". So I guess your name is ok.

PR #69, opened 2022-07-22
Finding duplicate filenames

DafinaTrufas:


palmskog:
Let's post the current full table here:

| Number of occurrences | Filename | Paths |
|:----------------------|:---------|:------|
| 2 | Common.v | ./theories/VLSM/Core/Equivocators/Composition/Common.v  ./theories/VLSM/Core/Equivocators/Common.v |
| 2 | LimitedEquivocation.v | ./theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/LimitedEquivocation.v  ./theories/VLSM/Core/Equivocation/LimitedEquivocation.v |
| 2 | Projections.v | ./theories/VLSM/Core/Equivocators/Projections.v  ./theories/VLSM/Core/Equivocators/Composition/Projections.v |

DUPLICATES.md
```diff
@@ -0,0 +1,5 @@
+| Number of occurrences | Filename | Paths |
```
palmskog, line 1: Can we make this into a task in `Makefile.coq.local` like for the ratios and `DUPLICATES.md`? As before, the problem with having this file checked in is that it will either become obsolete, or has to be updated/regenerated for each commit.

scripts/duplicates.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+BASE=$(dirname $0)
+echo "| Number of occurrences | Filename | Paths |"
+echo "|:----------------------|:---------|:------|"
+is_int() { test "$@" -eq "$@" 2> /dev/null; } 
+DUPLICATES=$(find . -iname "*.v" -type f -printf "%f\n" | sort | uniq -cd)
+for i in $DUPLICATES
+    do
+        echo  -n " $i "
+        find . -iname "$i" -exec $BASE/filenames.sh "{}" \; | sed 's/\t/\t | /g; s/^/| /g; s/$/ |/g'
```
palmskog, line 10: Please find some way to send command in `filenames.sh` **in line**, that is, without calling an external script file. This way, we will only have one file to maintain. The current functionality in `filenames.sh` is trivial, so we'd like to avoid having a file with it, if possible.

palmskog (changes requested):
This has the right idea, but please try to follow the pattern from the script on LOC ratios.

scripts/duplicates.sh
```diff
@@ -0,0 +1,15 @@
...
+        if ! is_int "$i"
+        then 
+            echo
+        fi
+    done
```
wkolowski, line 15:
```suggestion
#!/bin/sh
BASE=$(dirname $0)
echo "| Number of occurrences | Filename | Paths |"
echo "|:----------------------|:---------|:------|"
is_int() { test "$@" -eq "$@" 2> /dev/null; } 
DUPLICATES=$(find . -iname "*.v" -type f -printf "%f\n" | sort | uniq -cd)
for i in $DUPLICATES
do
  echo  -n " $i "
  find . -iname "$i" -exec $BASE/filenames.sh "{}" \; | sed 's/\t/\t | /g; s/^/| /g; s/$/ |/g'
  if ! is_int "$i"
  then 
    echo
  fi
done
```
Keep the script more compact - use 2 spaces for indentation instead of 4.

wkolowski (changes requested):
Besides addressing Karl's suggestions, please write some comments for `duplicates.sh`, so that people will have easier time when there's a need to revisit it.

palmskog (commented):
The code and Makefile task looks great now! 

But can you please remove the file `DUPLICATES.md` from the branch? Recall that we want to generate this file on command, rather than have it checked into version control.

DafinaTrufas:
> The code and Makefile task looks great now!
> 
> But can you please remove the file `DUPLICATES.md` from the branch? Recall that we want to generate this file on command, rather than have it checked into version control.

Oh, yes, I've forgotten to remove it.

scripts/duplicates.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+BASE=$(dirname $0)
+echo "| Number of occurrences | Filename | Paths |"
+echo "|:----------------------|:---------|:------|"
```
wkolowski, line 4: I think printing right at the start of the script is not a good idea. Please move it a bit down, to right before the loop. I think it makes more sense that way.

scripts/duplicates.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/sh
+BASE=$(dirname $0)
```
wkolowski, line 2: This variable is not used below, what's the point of it?

scripts/duplicates.sh
```diff
@@ -0,0 +1,15 @@
...
+echo "| Number of occurrences | Filename | Paths |"
+echo "|:----------------------|:---------|:------|"
+is_int() { test "$@" -eq "$@" 2> /dev/null; } 
+DUPLICATES=$(find . -iname "*.v" -type f -printf "%f\n" | sort | uniq -cd) # a list of the duplicate filenames and their frequencies
+for i in $DUPLICATES
+  do
+    echo  -n " $i " # the number of occurences / the filename
```
wkolowski, line 9: The comment is confusing, especially the `/` part in it.

scripts/duplicates.sh
```diff
@@ -0,0 +1,15 @@
...
+    find . -iname "$i" -exec echo -n $1 {} \; | sed 's/\t/\t | /g; s/^/| /g; s/$/ |/g' # the paths to the files with the current name
+    if ! is_int "$i" # if the value of variable i at the current iteration is a string (filename), move on a new line
```
wkolowski, line 11: This is confusing, you write `! is_int`, which reads like "is not int", but then you say you're testing whether `$i` is a string/filename.

Can you refactor `is_int` to something like `is_filename`?

wkolowski (changes requested):
I'd like to see a few more changes, and don't forget to delete `DUPLICATES.md`.

scripts/duplicates.sh
DafinaTrufas: I was initially using it and I've forgotten to remove it after I made some changes.

scripts/duplicates.sh
DafinaTrufas: By "is not int", I meant "is string". But yes, I will rename the function as you suggested.

DafinaTrufas: By "is not int", I meant "is string" (which means that it's a filename). The function tests whether the value of a variable is of type int, so I don't think that renaming it to "is_filename" would work.

DafinaTrufas: I tried to fix it, by making that comment a bit more explicit. Please tell me if now it's all right!

wkolowski: @DafinaTrufas

I meant that you should define a function `is_filename` that would be equivalent to `! is_int ...`.

Then you could write `if is_filename "$i"`, which would be much more readable. Is this possible in bash?

DafinaTrufas: I don't know, but I will search for it.

wkolowski (changes requested):
I think the rendering of the table is broken. There is no pipe `|` --between the number of occurrences and the filename-- before the number of occurrences, right at the start of a new line. Also, newline seems to be missing from the end. Please fix this.

wkolowski:
@DafinaTrufas

Also, I think the file paths are concatenated together. Could you make sure they are separated with `<br>`? (This will make them appear on separate lines).

DafinaTrufas:
> 
the problem is that, initially, I've fixed the table rendering by manually editing the Markdown file. And now, I don't manage to fix it by commands in the script.

scripts/duplicates.sh
```diff
@@ -0,0 +1,15 @@
...
+DUPLICATES=$(find . -iname "*.v" -type f -printf "%f\n" | sort | uniq -cd) # a list of the duplicate filenames and their frequencies
+for i in $DUPLICATES
+  do
+    echo  -n " | $i " # the number of occurences / the filename
+    find . -iname "$i" -exec echo -n "$1" {} \; | sed 's/\t/\t | /g; s/^/| /g; s/$/ |/g' # the paths to the files with the current name
+    if ! is_int "$i" # if the value of variable i at the current iteration is a string (filename), move on a new line 
+    then 
+      echo
+    fi
+  done
```
wkolowski, line 15: Looks like the final newline is missing.

theories/VLSM/Lib/TopSort.v
```diff
@@ -756,7 +756,7 @@ Context
   `{!StrictOrder precedes}
   .
 
-#[local] Lemma Forall_True : forall l : list A, Forall (fun _ => True) l.
+Lemma Forall_True : forall l : list A, Forall (fun _ => True) l.
```
wkolowski, line 759: You should not delete this. @palmskog suggested doing this only for the purpose of making `axiom-diagnostics.sh` work, but I am now taking care of that. Please remove this change.

wkolowski (changes requested):
Looks like you got some kind of reversion, as the table header is back at the top, where you previously put it right above the loop.

Also, I see you deleted the `#[local]` attribute from `Forall_True`. Looks like your commit got polluted from your axiom diagnostics changes.

scripts/duplicates.sh
wkolowski, line 2: It seems you deleted this line, but it's back there.

wkolowski (approved):
I think it is ok now.

Please don't merge, I want to do a final refactoring.

DafinaTrufas:
> I think it is ok now.
> 
> Please don't merge, I want to do a final refactoring.

Sure! I wait for you to tell me when I can merge.

wkolowski (approved):

palmskog (approved):

PR #72, opened 2022-07-25
The lemma `elem_of_list_split_2`

wkolowski:
This is needed for UMO.


theories/VLSM/Lib/StdppExtras.v
```diff
@@ -324,3 +324,25 @@ Proof.
...
   by apply drop_S.
 Qed.
+
+(**
+  When a list contains two elements, either they are equal or we can split
+  the list into three parts separated by the elements (and this can be done
+  in two ways, depending on the order of the elements).
+*)
+Lemma elem_of_list_split_2 :
+  forall {A : Type} {x y : A} {l : list A},
```
palmskog, line 334: Do we really want to make *all* arguments implicit here? I think it makes more sense to have the list and elements as regular arguments.
```suggestion
  forall {A : Type} (l : list A) x y,
```

palmskog (approved):
Fine, but see suggestion.

theories/VLSM/Lib/StdppExtras.v
wkolowski: These should be inferable from `x \in l` and `y \in l`... but I'll change it as it doesn't matter anyway.

PR #73, opened 2022-07-25
Fix top-level indentation

wkolowski:
I changed all top-level stuff (`Definition`, `Lemma`, `Section` etc.) to start right at the beginning of the new line, without any indentation. This makes the style more uniform across the project, but also has the added benefit of making the axiom diagnostics script work better.


traiansf (approved):
I agree with this, and I'm actually glad to see it happen; however, since it's a huge change, I would suggest that you please wait a bit before merging (say, until tomorrow morning), to see whether there is any argument against it.

DafinaTrufas (approved):

palmskog (approved):
I agree with the general idea of having everything with no indentation by default (this is what MathComp uses). However, I will have to trust that this is all this PR does, since it's just so much code that changes that it takes too much time check everything.

wkolowski:
@palmskog Besides fixing indentation I added or removed a few newlines here and there. Maybe this is why the diff looks so terrible...

PR #74, opened 2022-07-26
Renaming files 

DafinaTrufas:
I renamed the files which had the same name (according to Traian's suggestions) and replaced with the new filenames also in _CoqProject and in the Requires. 
I cannot figure out why I get the following error, when running make: 
Cannot find a physical path bound to logical path
Core.Equivocators.Composition.EquivocatorsCompositionProjections with prefix VLSM


wkolowski (changes requested):
It fails because of a cyclic dependency, which must have been erroneously introduced during the renaming.

wkolowski (approved):

palmskog (approved):
Looks good, please merge whenever convenient.

PR #76, opened 2022-07-28
Porting axiom diagnostics script from casper-cbc-proofs

wkolowski:


palmskog (approved):
This is fine, but do we really need to document it in the README here? I think it suffices to have it documented elsewhere, since the main use is by the development team and not external users. So I think we can roll back the updates to `README.md` and `meta.yml`.

traiansf:
Still, it might be good to have a file describing the various scripts and the ways to employ them. Maybe a `README.md` file in the scripts directory?

PR #77, opened 2022-07-28
Purge the Preamble from unused and useless definitions and lemmas

wkolowski:
Note: this will require 2-3 one-line changes in casper-cbc-proofs.


traiansf (approved):
I have nothing against it.

theories/VLSM/Lib/Preamble.v
```diff
@@ -219,175 +219,6 @@ Proof.
...
-Definition noneOrAll : option Prop -> Prop := default True.
-
-(* https://coq.discourse.group/t/writing-equality-decision-that-reduces-dec-x-x-for-opaque-x/551/2 *)
-
-Lemma eq_dec_refl A (eq_dec : forall x y : A, {x = y} + {x <> y}) x :
-  eq_dec x x = left eq_refl.
-Proof.
-  destruct (eq_dec x x) as [| []]; [| done].
-  by rewrite K_dec_type with (P := fun prf => prf = eq_refl).
-Qed.
```
palmskog, line 235: This looks like a useful lemma to have around - it's nontrivial for most Coq users, and can be useful when reasoning about equality, as shown in the forum link. So I think both the lemma and the link should stay.

theories/VLSM/Lib/Temporal.v
```diff
@@ -144,18 +144,18 @@ Proof.
   intro Hprogress.
   apply Classical_Prop.NNPP.
   intro H.
-  pose proof (not_ex_all_not _ _ H); clear H. simpl in H0.
   assert (forall x, Forever (Eventually (fun s => hd s <> x)) s).
   {
     intro x.
+    assert (∀ n : A, ¬ Eventually (Forever (λ s : Stream A, hd s = n)) s) by firstorder.
```
palmskog, line 150: I don't the see the need to use UTF8 here.
```suggestion
    assert (forall n : A, ~ Eventually (Forever (fun s : Stream A, hd s = n)) s) by firstorder.
```

palmskog (changes requested):
Looks good in general, but please address comments.

theories/VLSM/Lib/Temporal.v
wkolowski: Hmm, I think they got there by copy-pasting.

palmskog (approved):

PR #79, opened 2022-08-02
Removing lemmas

DafinaTrufas:


wkolowski (commented):
So far so good, but is it the final version of this PR or are you going to add more commits?

If it's not final, then it's a good idea to mark it as a "draft PR" and only request a review when it's ready to be merged.

DafinaTrufas:
It's not final, I'm going to add more commits. Sorry, I wasn't aware of the possibility to mark a PR as draft.

theories/VLSM/Lib/TraceClassicalProperties.v
```diff
@@ -16,22 +16,22 @@ Context {A B : Type}.
 
 (** ** Relating finiteness and infiniteness *)
 
-Lemma not_infiniteT_finiteT : forall tr : trace,
+(*Lemma not_infiniteT_finiteT : forall tr : trace,
```
palmskog, line 19: @DafinaTrufas the lemmas in this file have been chosen specifically to provide a complete trace theory. They shouldn't be removed, even if the project still compiles. I'd advise to focus on lemmas in utility files like `*Extras.v`.

DafinaTrufas: Ah, ok, I didn't know what is meant by utility files. So the lemmas from the other files above shouldn't be removed neither?

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1301,10 +1301,10 @@ Fixpoint complete_prefix
                                end
   end.
 
-Example complete_prefix_some : complete_prefix [1;2;3;4] [1;2] = Some [3;4].
-Proof. itauto. Qed.
-Example complete_prefix_none : complete_prefix [1;2;3;4] [1;3] = None.
-Proof. itauto. Qed.
+(*Example complete_prefix_some : complete_prefix [1;2;3;4] [1;2] = Some [3;4].
```
traiansf, line 1304: It might be a good idea not to remove examples if you don't remove the definitions they are examples for, as their purpose is to serve as documentation for usage patterns.
@palmskog what do you think?

traiansf:
I feel a little biased about removing some results about streams and stream filtering and also about topological sorting (maybe because I've been working on them).

My feeling is that some are general purpose lemmas, which might not be used now, but could be eventually (e.g., about streams / stream filtering), while others help make the argument about the correctness of an algorithm (e.g., about topological sorting).

On the other hand, we don't really know when we're going to need those and might not make sense to maintain them in the hope that they would sometime be useful. So I'm wondering whether it would make sense to tag the current master and remove them, keeping it mind that if we sometime need them, we can take a look at the tagged version.

theories/VLSM/Lib/ListExtras.v
palmskog: @traiansf @DafinaTrufas I completely agree that we should leave examples in. The main goal here in my view should be to find "utility" lemmas that don't express anything deep and don't validate any other theorem.

palmskog (changes requested):
@DafinaTrufas an important goal with removing lemmas is to remove *code*, that is, to make files smaller and more surveyable. Hence, you should remove lemmas and not comment them out.

However, that is not to say that all currently commented-out lemmas in this PR should actually be removed in the end. But please change the PR to use full removal, and we'll take it from there.

theories/VLSM/Lib/TraceClassicalProperties.v
palmskog: @DafinaTrufas I mean the you should roll back removals/comment-outs in all `Trace`-related files in `VLSM/Lib`.

DafinaTrufas:
> @DafinaTrufas an important goal with removing lemmas is to remove _code_, that is, to make files smaller and more surveyable. Hence, you should remove lemmas and not comment them out.
> 
> However, that is not to say that all currently commented-out lemmas in this PR should actually be removed in the end. But please change the PR to use full removal, and we'll take it from there.

Ok. I'm changing it now.

theories/VLSM/Lib/SsrExport.v
```diff
@@ -4,4 +4,4 @@ From Coq Require Export ssreflect.
 #[export] Set SsrOldRewriteGoalsOrder.
 #[export] Set Asymmetric Patterns.
 #[export] Set Bullet Behavior "None".
-Definition iYC2 := True.
```
palmskog, line 7: Please roll back this removal. This is needed for our documentation toolchain to work.

wkolowski (changes requested):
You removed a lemma I recently added (`elem_of_list_split_2`), which will break casper-cbc-proofs. You should certainly add it back.

Regarding the other changes, I have mixed feelings. I think removing lemmas that refer to `StrictlyComparable` or other in-house stuff seems fine, but these extras that refer to `EqDecision` etc. seem pretty general and useful.

Maybe we should try to submit these lemmas to the standard library and stdpp before we remove them?

Also, I wonder: does the script consider usage cross-project? Removing `elem_of_list_split_2` is clearly a false positive, as it is used in casper-cbc-proofs.

theories/VLSM/Lib/StdppExtras.v
```diff
@@ -712,25 +628,3 @@ Proof.
   rewrite contra in H0.
   destruct H0; inversion H0.
 Qed.
-
-(**
-  When a list contains two elements, either they are equal or we can split
-  the list into three parts separated by the elements (and this can be done
-  in two ways, depending on the order of the elements).
-*)
-Lemma elem_of_list_split_2 :
```
palmskog, line 721: This was recently added by @wkolowski for an as-yet-unmerged PR. Please reinstate.

theories/VLSM/Lib/TopSort.v
```diff
@@ -135,21 +135,6 @@ Proof.
   by destruct (decide (precedes a a)).
 Qed.
 
-Lemma precedes_asymmetric
```
palmskog, line 138: I remember discussing these lemmas with @traiansf, and they are very useful to give a complete theory about topological sorting. Please roll back changes to `TopSort.v`.

theories/VLSM/Lib/StdppExtras.v
wkolowski: No, it's been merged for quite some time already.

wkolowski (commented):
I see my suggestions were addressed, but I still don't know whether these changes are for good. I will let Karl decide.

palmskog:
There are still too many removals here that require more discussion. I think we should trim the PR to only the uncontroversial removals. I may be able to get to this next week.

theories/VLSM/Lib/SsrExport.v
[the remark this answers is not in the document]
palmskog: @wkolowski some of the tooling (coq-dpdgraph, I think) requires there to be a constant defined in each file. So we add this dummy constant.

wkolowski:
@palmskog It seems this PR has been stale for around 1.5 years and also I have removed plenty of duplicated/unnecessary lemmas over time, e.g. some of the stuff in `ListExtras.v` and almost everything from `RealsExtras.v`. Should this PR be closed?

PR #83, opened 2022-08-04
Adding the aixom-analysis documentation and the CODING_CONVENTIONS.md

DafinaTrufas:


TOOLS.md
```diff
@@ -43,3 +43,20 @@ In this single file case, all the project's `-Q`/`-R` options must be manually p
 ```
 
 Note that the results of running the script must be manually validated. If the script removes dependencies which are actually used, they will need to be added back manually.
+
+### Axiom use analysis
```
palmskog, line 47: Since this is not related to coq-tools, please use another level of headers:
```suggestion
## Axiom use analysis
```

palmskog (approved):
Looks good, but please fix the header level as indicated.

palmskog (changes requested):
Looking closer, I don't see the `CODING_CONVENTIONS.md` document, please add it before merging.

palmskog (approved):

PR #84, opened 2022-08-08
Fixing warnings related to Instance locality

DafinaTrufas:


palmskog:
Looks reasonable, but perhaps @traiansf could comment on whether we want `#[export]` for some of these classes, or something more restrictive like `#[local]` is enough.

palmskog (commented):
I'm happy with this PR, but perhaps Traian could comment, and if he has no problem then we merge.

wkolowski (approved):

theories/VLSM/Lib/Preamble.v
```diff
@@ -665,13 +665,13 @@ Proof.
 Qed.
 
 (* Now we can have the following for free : *)
-Instance ComposeStrictlyComparable (X Y : Type) `{StrictlyComparable X} `{StrictlyComparable Y} : StrictlyComparable (X * Y) :=
+#[export] Instance ComposeStrictlyComparable (X Y : Type) `{StrictlyComparable X} `{StrictlyComparable Y} : StrictlyComparable (X * Y) :=
```
traiansf, line 668: I wonder whether this and the next are still used; if they are only used for CBC, maybe we could move them there and make them local? Not necessarily as part of this PR.

traiansf (approved):

PR #85, opened 2022-08-08
Axiom Diagnostics: option for keeping temporary file

wkolowski:
I tested it and it works for me...


palmskog (approved):
I think there are less intrusive ways of passing a parameter with a default value (something like `KEEP_TMP =? true`), but let's go with this for now.

PR #87, opened 2022-08-12
A tactic which automates reasoning about the state_update function

wkolowski:
I noticed that there are some clear patterns when reasoning about the `state_update` function from `Composition.v`. Namely, if the indices are equal, we `rewrite state_update_eq`, when they are not we `rewrite state_update_neq` and when we don't know, we do `destruct (decide (i = j)); subst` and then we `rewrite` accordingly.

This often causes the nesting level of the proof to increase by 1, which makes proofs less readable. So maybe it's a good idea to have a tactic for this?

### DIFF
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -843,7 +843,7 @@ End sec_composite_vlsm.
 End VLSM_composition.
 
 (**
-  A nice little tactic for dealing with [state_update].
+  Hint database and tactic for dealing with updates and lifting via [state_update].
 *)
 
 Create HintDb state_update.
@@ -857,9 +857,6 @@ Create HintDb state_update.
 
 Ltac state_update_simpl :=
   autounfold with state_update in *; autorewrite with state_update in *.
-
-Ltac state_update i j :=
-  destruct (decide (i = j)); subst; state_update_simpl.
 
 (**
    These basic projection lemmas relate
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -1675,7 +1675,7 @@ Proof.
   rewrite Heq_s in Heqs at 1; clear Heq_s.
   specialize (unique_transition_to_state Ht1 Ht2) as Heq;
     destruct_and! Heq; subst; repeat split.
-  extensionality j; destruct (decide (i = j)); [by subst |].
+  extensionality j; state_update i j; [done |].
   apply f_equal with (f := fun s => s j) in Heqs.
   by state_update_simpl.
 Qed.
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -843,6 +843,27 @@ End sec_composite_vlsm.
 End VLSM_composition.
 
 (**
+  A nice little tactic for dealing with [state_update].
+*)
+Ltac state_update_simpl :=
+repeat match goal with
+| |- context [state_update _ _ ?i _ ?i] => rewrite (state_update_eq _ _ i)
+| Heq : ?i <> ?j |- context [state_update _ _ ?i _ ?j] =>
+  rewrite (state_update_neq _ _ i _ j) by done
+| Heq : ?j <> ?i |- context [state_update _ _ ?i _ ?j] =>
+  rewrite (state_update_neq _ _ i _ j) by done
+| H : context [state_update _ _ ?i _ ?i] |- _ => rewrite (state_update_eq _ _ i) in H
+| Heq : ?i <> ?j, H : context [state_update _ _ ?i _ ?j] |- _ =>
+  rewrite (state_update_neq _ _ i _ j) in H by done
+| Heq : ?j <> ?i, H : context [state_update _ _ ?i _ ?j] |- _ =>
+  rewrite (state_update_neq _ _ i _ j) in H by done
+(*   | |- context [state_update _ _ ?i _ ?j] => destruct (decide (i = j)); subst *)
+end.
+
+Ltac state_update i j :=
+  destruct (decide (i = j)); subst; state_update_simpl.
+
+(**
    These basic projection lemmas relate
    the [valid_state_prop] and [input_valid_transition] of
    a composite VLSM back to those conditions holding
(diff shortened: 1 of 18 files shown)


palmskog (changes requested):
I'm sympathetic to the idea of domain-specific rewriting automation. However, the custom tactic is basically just applying rewrites. A more disciplined and maintainable way to do this is to use *named* rewriting databases (`Hint Rewrite my_eq : my_db`) and for example the [autorewrite](https://coq.inria.fr/refman/proofs/automatic-tactics/auto.html#coq:tacn.autorewrite) tactic.

wkolowski:
@palmskog Yes, `autorewrite` seems a bit better here, as in later modules there are some new lemmas about particular incarnations of `state_update`. I will refactor this refactoring to use it, but it can make this refactoring take a lot more time.

palmskog:
@wkolowski since this looks like pure refactoring PR (no new results), it's fine by me if it takes time (and it shouldn't affect anything else). You might consider marking the PR as a WIP until you change it to use rewriting databases.

theories/VLSM/Core/Composition.v
```diff
@@ -842,6 +842,25 @@ End sec_composite_vlsm.
 
 End VLSM_composition.
 
+(**
+  A nice little tactic for dealing with [state_update].
```
palmskog, line 846:
```suggestion
  Hint database and tactic for dealing with updates and lifting via [state_update].
```

theories/VLSM/Core/Composition.v
```diff
@@ -842,6 +842,25 @@ End sec_composite_vlsm.
...
+
+#[export] Hint Unfold lift_to_composite_state : state_update.
+#[export] Hint Unfold lift_to_composite_state' : state_update.
+
+Ltac state_update_simpl :=
+  autounfold with state_update in *; autorewrite with state_update in *.
+
+Ltac state_update i j :=
+  destruct (decide (i = j)); subst; state_update_simpl.
+
```
palmskog, line 863: I don't see the point of introducing this tactic. It basically just composes two standard tactic calls with an automation tactic. I think this only obscures what is going on in proofs and will make them harder to maintain.
```suggestion
```

theories/VLSM/Core/Composition.v
```diff
@@ -1660,9 +1673,9 @@ Proof.
   rewrite Heq_s in Heqs at 1; clear Heq_s.
   specialize (unique_transition_to_state Ht1 Ht2) as Heq;
     destruct_and! Heq; subst; repeat split.
-  extensionality j; destruct (decide (i = j)); [by subst |].
+  extensionality j; state_update i j; [done |].
```
palmskog, line 1676: Here is a case in point of why the `state_update` tactic is more trouble than it's worth: instead of a regular case analysis with a substitution, we are now running a powerful and possibly long-running tactic under the hood.

palmskog (changes requested):
The rewrite database and the `state_update_simpl` tactic is good. But I completely disagree about the `state_update` tactic. In particular, several easy proofs were given *more* automation.

theories/VLSM/Core/Composition.v
wkolowski: Ok, maybe you are right, especially that `state_update_simpl` is not a succeed-or-fail kind of tactic. I'll remove `state_update i j`.

theories/VLSM/Core/Equivocators/Composition/EquivocatorsCompositionProjections.v
```diff
@@ -28,6 +28,8 @@ Context {message : Type}
   (PreFree := pre_loaded_with_all_messages_vlsm Free)
   .
 
+Hint Unfold equivocator_descriptors_update : state_update.
```
palmskog, line 31: Please ensure that all new `Hint` have a locality (probably `#[export]` for all of them).
```suggestion
#[export] Hint Unfold equivocator_descriptors_update : state_update.
```

theories/VLSM/Core/Equivocators/Equivocators.v
```diff
@@ -446,6 +446,23 @@ Proof.
   lia.
 Qed.
 
+Hint Rewrite @equivocator_state_update_size : state_update.
```
palmskog, line 449: It worries me a lot to see a database with 20+ rewriting lemmas, this can lead to significant slowdowns with `autorewrite` (recall that it tries *everything* repeatedly). How about instead creating different databases for each "type" of update, such as `equivocator_state_update`?

palmskog (changes requested):
This is a clear step in the right direction, but at the minimum, all `Hint`s must have locality. I also worry a lot about the size of rewrite databases.

theories/VLSM/Core/Equivocators/Composition/EquivocatorsCompositionProjections.v
wkolowski: Umm, this is inside a `Section`, so I guess your locality is incorrect here. If I recall correctly, all `Hint`s I added outside `Section`s are `#[export]`, and the ones inside sections have no locality given.

palmskog: But if it's indeed inside a section, you should add `#[local]` to indicate that. The point is that we want explicit locality for all `Hint`, no matter if it's `#[export]` or not.

wkolowski: Ok, I'll add them.

theories/VLSM/Core/Equivocators/Equivocators.v
wkolowski: 14 of these 20+ are a matter of a single file (`Equivocators.v`), so it's probably less bad than it sounds. Also, from my experience, there are maybe 1 or 2 places where there was some slowdown, but these were chains of 10 or 20 tactics which probably should be refactored anyway.

Also, I think the separation might not be that easy to achieve, since these various flavors of `state_update` are somewhat interdependent. Maybe we should just have a separate tactic for `equivocator_state_update` if that's possible?

palmskog: Yes, the main issue might be with equivocator state updates. I think a separate `equivocator_state_update_simpl` tactic might be a good idea, to separate at least to some extent by the domain. Recall that one can configure `autorewrite` with multiple databases.

palmskog (approved):
I think it all looks reasonable now, even if the duplicated hints (local vs export) are a bit unfortunate.

theories/VLSM/Core/Equivocators/Equivocators.v
```diff
@@ -446,6 +446,27 @@ Proof.
   lia.
 Qed.
 
+#[local] Hint Rewrite @equivocator_state_update_size : equivocator_state_update.
```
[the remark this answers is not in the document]
wkolowski: Right, the manual proofs look shorter than the `Hint` base needed to automate them.

traiansf:
A quite nice and elegant solution. Thanks @wkolowski  and @palmskog for making it happen.

PR #88, opened 2022-08-12
Refactoring of the Preamble

wkolowski:


palmskog (commented):
This PR mixes many different kinds of changes, which makes it hard to review. I'm prepared to sign off on the changes related to dependent pairs. But I'm not so sure about the other changes and ideally @bmmoore should sign off on them instead of me.

So if you want the dependent pair changes merged quickly @wkolowski you could break them off into another PR and I will approve that PR.

theories/VLSM/Lib/Preamble.v
```diff
@@ -269,19 +218,20 @@ Proof.
...
 Definition strict_maximal_among `(R : relation A) := strict_minimal_among (flip R).
 
-(* Reflexivity of comparison operators *)
+(** ** Comparison operators *)
+
+(** *** Reflexive comparison operators *)
+
 Class CompareReflexive {A} (compare : A -> A -> comparison) : Prop :=
-    compare_eq : forall x y, compare x y = Eq <-> x = y.
+  compare_eq : forall x y, compare x y = Eq <-> x = y.
```
bmmoore, line 231: I see it's not new, but this name is bad. Reflexivity would be something more like `forall x, compare x x = Eq`. Merely being reflexive doesn't prevent there being "equivalence classes" with xRy and yRx, as long as also xRx and yRy. This seems more like an antisymmetric relation.

That assumes we apply relation terminology to a compare function by thinking that `compare` represents a relation `R` in the sense that `match compare x y with Lt => R x y /\ ~R y x | Eq => ~R x y /\ ~R y x | Gt => ~R x y /\ R y x end` is always true.

theories/VLSM/Lib/Preamble.v
```diff
@@ -568,125 +513,109 @@ Proof.
...
+  (X : Type) {Xsc : StrictlyComparable X} : StrictlyComparable (option X) :=
+{
+  inhabited := None;
+  compare := option_compare compare;
+  compare_strictorder := strictorder_option Xsc;
+}.
+
+(** *** Comparison for pairs *)
+
+Definition compare_compose
```
bmmoore, line 526: Again complaining about a preexisting name. This one isn't too bad but "compose" is not very specific, and might bring to mind relation composition. How about compare_pair or maybe compare_lex_pair (I can't think of another sensible way to compose strict orders to get a strict order on pairs, besides reverse lexicographic)

bmmoore (approved):
Looks good to me, except for complaints about preexisting names.
I would "request changes" over renaming CompareReflexive if that was new.
compare_compose is just vague, not misleading.

theories/VLSM/Lib/Preamble.v
wkolowski: Yeah, the right-to-left direction of this corresponds to reflexivity, whereas left-to-right says that equality is decidable.

I will leave renamings to a future PR.

theories/VLSM/Lib/Preamble.v
wkolowski: I agree that the name is bad and it should be either `compare_pair` or `compare_lex(icographic)`, but I will leave renamings for a future PR.

PR #89, opened 2022-08-17
Removing trailing whitespace using the shell script

DafinaTrufas:


scripts/remove-whitespace.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+find . -iname "*.v" -exec sed -i -e "s/[[:blank:]]*$//" "{}" \;
```
wkolowski, line 2: Please add a newline at the end.

wkolowski (approved):
Looks ok, but I didn't triple-check precisely how the regex works.

scripts/remove-whitespace.sh
DafinaTrufas: Yes, indeed, I will open a seaparate PR for the newline-deduplication task.

palmskog (approved):
Looks good to me.

PR #90, opened 2022-08-18
Striping newlines

DafinaTrufas:


scripts/strip_newlines.py
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env python
+import sys
+import os
+
+sys.path.insert(0, '/home/dafina/vlsm/coq-tools')
```
palmskog, line 5: This looks worrying for other people using the script. Are you sure we can't directly call some `coq-tools` script directly from the shell?

At least, this should be changed from a hard-coded absolute path, e.g., you could switch to `../coq-tools` or similar. And please add some documentation on how to run the script in `TOOLS.md`.

palmskog (changes requested):
The changes look good, but please consider the comments about the script (and documentation).

scripts/strip_newlines.py
DafinaTrufas: I am not sure that it couldn't be done using directly a script from `coq-tools`. Initially, I tried it, but I didn't manage that way.

DafinaTrufas:
Is this form of the script acceptable?

scripts/strip_newlines.py
```diff
@@ -0,0 +1,32 @@
...
+def getListOfFiles(dirName):
+    listOfFile = os.listdir(dirName)
+    allFiles = list()
+    for entry in listOfFile:
+        fullPath = os.path.join(dirName, entry)
+        if os.path.isdir(fullPath):
+            allFiles = allFiles + getListOfFiles(fullPath)
+        else:
+            allFiles.append(fullPath)
+    return allFiles
```
wkolowski, line 17: This function looks very generic - it finds all files in a directory and recursively in its subdirectories. I would expect this kind of function to be provided by the library you are using. Did you check if `os` or `sys` already provides such a function?

Also, it's not already provided by the library, isn't there a more functional way of doing it? Like, using `map` and `filter`?

scripts/strip_newlines.py
```diff
@@ -0,0 +1,32 @@
...
+        fullPath = os.path.join(dirName, entry)
+        if os.path.isdir(fullPath):
+            allFiles = allFiles + getListOfFiles(fullPath)
+        else:
+            allFiles.append(fullPath)
+    return allFiles  
+
+
+def main():
+    dirName = './theories/VLSM'
```
wkolowski, line 21: This hard-coded path makes the script not portable across repos, as `casper-cbc-proofs` has `CBC`, `CBC_VLSM` and `VLSM_index` in the `theories/` directory.

```suggestion
    dirName = './theories'
```

scripts/strip_newlines.py
```diff
@@ -0,0 +1,32 @@
...
+    listOfFiles = getListOfFiles(dirName)
+    for f in listOfFiles:
+        if f.endswith(".v"):
+            with open (f, 'r') as file:
+               content = file.read()
+            with open (f, 'w') as file:
+               file.write(strip_newlines(content, int(sys.argv[1])))
```
wkolowski, line 28: What is `sys.argv[1]` supposed to contain? If you're using command line arguments, you should parse them at the beginning of the script/main function.

wkolowski (changes requested):
You should change the hard-coded path and also look for a more compact way of writing the script, if it's possible.

scripts/strip_newlines.py
wkolowski: Go do something like `max_consecutive_newlines = sys.argv[1]` at the beginning of the main function.

Also, what happens when the argument can't be parsed as a positive natural number?

DafinaTrufas: If max_consecutive_newlines is a negative number, the content should be returned unchanged.

wkolowski: And what if the argument is not a number at all?

DafinaTrufas: The `max-consecutive-newlines` it's taken as argument by the `split-newlines` function, which is defined in `coq-tools`, and in that function are handled the cases when `max-consecutive-newlines` it's negative, zero, or positive. The case when it's not a number is not handled there. Do you mean I should return some error message in this case?

wkolowski: Yes, I you should output some nice error message like "not a number", so that the user doesn't get a Python exception thrown at him.

scripts/strip_newlines.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+def main():
+    dirName = './theories/VLSM'
+    listOfFiles = getListOfFiles(dirName)
+    for f in listOfFiles:
+        if f.endswith(".v"):
+            with open (f, 'r') as file:
+               content = file.read()
+            with open (f, 'w') as file:
```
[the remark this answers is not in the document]
wkolowski: @DafinaTrufas What about something like `with open(f, 'rw') as file:`?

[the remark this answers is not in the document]
DafinaTrufas: I used `with open(f, 'r+')` (the changes are in PR#93). Is this way also all right?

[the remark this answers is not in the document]
wkolowski: Sorry, I didn't see that PR when I clicked "Unresolve". I think it's ok.

wkolowski (approved):

palmskog (approved):

PR #93, opened 2022-08-19
Refactoring the strip-newlines.py script and adding some documentation for it

DafinaTrufas:


DafinaTrufas:
Now I think the only review I didn't address at all is the one regarding the case when the `max-consecutive-newlines` argument isn't a number. I'm not sure how to deal with this case.

scripts/strip_newlines.py
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env python
+import sys
+import os
+
+sys.path.insert(0, './coq-tools')
+from strip_newlines import strip_newlines
+
+
```
wkolowski, line 8: It's a bit ironic that you have double newlines in a script supposed to remove them :)

scripts/strip_newlines.py
```diff
@@ -0,0 +1,23 @@
...
+def main():
+    max_consecutive_newlines = int(sys.argv[1])
```
wkolowski, line 10: If I do `int("asd")` in Python 3 shell, I get an exception:

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 10: 'asd'
```

The exception can be caught and handled more gracefully:

```python
try:
  int("asd")
except ValueError:
  print("The argument must a positive number!")
```

Read https://docs.python.org/3/tutorial/errors.html to learn more.

wkolowski (changes requested):

scripts/strip_newlines.py
DafinaTrufas: Oh, I will fix that.

scripts/strip_newlines.py
DafinaTrufas: Ok, I'll handle this exception.

wkolowski (approved):

TOOLS.md
```diff
@@ -60,3 +60,15 @@ root directory:
 ```shell
 make axioms path=theories/CBC
 ```
+
+## Striping newlines
```
wkolowski, line 64: I think you made a typo there: https://grammarist.com/usage/stripedstriping-vs-strippedstripping

TOOLS.md
```diff
@@ -60,3 +60,15 @@ root directory:
 ```shell
 make axioms path=theories/CBC
 ```
+
+## Striping newlines 
+
+To strip consecutive newlines exceeding a certain number, run the `strip-newlines.py` script, passing as command line argument `max + 1` (where max is the maximum number of consecutive newlines you want to accept). 
+
+In case you pass a negative number as argument, the content will remain unchanged, and if the argument you pass is `0`, then all strings of consecutive newlines will be replaced by a single space.
```
wkolowski, line 68: You meant "newline" there, I guess?

```suggestion
In case you pass a negative number as argument, the content will remain unchanged, and if the argument you pass is `0`, then all strings of consecutive newlines will be replaced by a single newline.
```

TOOLS.md
```diff
@@ -60,3 +60,15 @@ root directory:
...
+
+For example, if you want to limit the number of consecutive newlines to 1, run:
+
+```shell
+python scripts/strip-newlines.py 2
```
wkolowski, line 73: I think this is confusing. The argument is 2, but the effect is that there is at most one consecutive newline. Maybe refactor so that the argument is `max`, not `max + 1`?

scripts/strip_newlines.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python
```
wkolowski, line 1: Looks like the script is doubled. Please delete one of them.

wkolowski (changes requested):
Please address my suggestions. Also, what is this binary `.pyc` file?

TOOLS.md
DafinaTrufas: This information is taken from the `coq-tools` script which I imported to my script.

DafinaTrufas: And there was saying `space`, not `newline`.

TOOLS.md
DafinaTrufas: Again, I maintained the idea from the `coq-tools` script. This is the explanation from there:
```If max_consecutive_newlines > 0, all strings of consecutive
    newlines in excess of max_consecutive_newlines are replaced by
    max_consecutive_newlines newlines.```

DafinaTrufas:
Regarding the `.pyc` file, that is a binary file which is automatically generated for the `.py` file.

wkolowski:
@DafinaTrufas It's probably not a good idea to commit autogenerated binary files to the repo. Please remove it.

palmskog (approved):
This looks good to me now.

wkolowski (approved):
It's probably good enough now.

PR #95, opened 2022-08-19
Refactor a few typeclass instances in the Preamble

wkolowski:


theories/VLSM/Lib/Preamble.v
```diff
@@ -376,11 +370,7 @@ Class StrictlyComparable (X : Type) : Type :=
 }.
 #[global] Hint Mode StrictlyComparable ! : typeclass_instances.
 
-#[export] Instance strictly_comparable_eq_dec `{StrictlyComparable M} : EqDecision M.
-Proof.
-  intros x y.
-  apply compare_eq_dec.
-Qed.
+Coercion compare_strictorder : StrictlyComparable >-> CompareStrictOrder.
```
palmskog, line 373: I don't really see the motivation for adding this coercion. If we don't need for existing lemmas, why add it?

wkolowski: We do need the instance `compare_eq_dec` in a few places, but `strictly_comparable_eq_dec` is superfluous, since `StrictlyComparable` is a `CompareStrictOrder`, so it already has decidable equality, it's just a matter of find the right instance (`compare_eq_dec`). Without this coercion, this instance is not found.

palmskog: If the goal is indeed to get some instance, why not just a regular `#[export] Existing Instance field.`?

wkolowski: @palmskog Ok, I misremembered. There indeed are a few places (in `casper-cbc-proofs`) where the coercion is needed, and not just the instance. I could try refactoring that, but I think it's easier to keep the coercion. My goal for this refactoring was to get rid of these "doubled" instances, which seemed silly to me.

palmskog (approved):

PR #97, opened 2022-08-23
Replacing #[global] Instance with #[export] Instance

DafinaTrufas:


palmskog (approved):
Looks good.

scripts/replacing-global.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+BASE=$(dirname $0)
+find . -iname "*.v" -exec sed -i -e 's/#\[global\]\(.*\)Instance/#\[export\]\1Instance/' "{}" \;
```
wkolowski, line 3: Newline missing at the end.

Also, this script doesn't catch newlines between `#[global]` and `Instance`, right? Did you hunt these cases down manually?

wkolowski (approved):

scripts/replacing-global.sh
DafinaTrufas: Oh, I found here https://www.gnu.org/software/sed/manual/html_node/Regular-Expressions.html, that ". Matches any character, including newline."

wkolowski: The thing is, `sed` works line-by-line. So there's no newline to be matched.

wkolowski: But I did some `grep`ping and it seems you didn't miss anything.

DafinaTrufas: Ah, ok, I see. From what I've seen, there are no cases like the one you mentioned. Would it be sufficient to manually double check that there aren't, or should I find a way to include this check in the script?

wkolowski: No need to bother. `grep -h "\#\[global\]" $(find -name "*.v")` says it's all `#[global] Hint Mode...`, so you can go on and merge the PR.

wkolowski (changes requested):
It seems you got some kind of regression, as the `strip_newlines` script got deleted.

DafinaTrufas:
I deleted `strip_newlines.py`, because I had this script actually duplicate (`strip_newlines.py` and `strip-newlines.py`). I know you've also told me to remove one of the two, and I did it, but for some reasons today there were again both there, so I had to fix the problem.

wkolowski (approved):

PR #98, opened 2022-08-23
Remove a few comparison lemmas from the Preamble

wkolowski:
Note that `compare_eq_lt` (which proves  `~ compare x x = Lt`) is exactly the same thing as `compare_lt_irreflexive` (which proves `Irreflexive (compare_lt compare).`).

Similarly, `compare_lt_neq` is almost `compare_lt_irreflexive`.


palmskog (approved):
Looks reasonable.

PR #100, opened 2022-08-25
Refactor compare for pairs and options

wkolowski:


palmskog (commented):
The PR changes a lot of names of lemmas, but it's not clear to be what is gained at a high level. Can you give some intuition and/or overview of why the renaming is done?

wkolowski:
@palmskog I think @bmmoore complained in one of the previous reviews that existing names are sometimes bad, particularly for `compare_compose` (which I agree is a bad name for comparison of pairs). So I set out to rename `compare_compose` to `pair_compare`, together will all its lemmas. Then I noticed the naming is inconsistent across types (so lemmas for `option` had different name patterns than these for pairs), so I tried to make it more uniform.

palmskog (approved):
OK, it's not obvious how all the new names are an improvement, but there's nothing significant to complain about.

PR #101, opened 2022-08-25
Unbundle the inhabited field from the StrictlyComparable typeclass

wkolowski:


palmskog (approved):
This looks like a simplification, so fine by me, but please ensure it works with clients of the repo before merging.

PR #102, opened 2022-08-27
Cleanup uses of the specialize tactic

wkolowski:


traiansf:
Why is it important to remove usages of the specialize tactic?

palmskog:
Same question here, I don't really understand at all why `specialize` and `spec` should be bad.

wkolowski:
@traiansf I think most uses of `specialize` in our codebase are code smells (proof smells?):
- Usually `apply` and a backward style of proof are better than `specialize` and forward style.
- Sometimes the place of specialization is far from the place of use, which makes understanding the proof harder.
- There are even some `specialize`s which are completely unnecessary, which is also confusing.
- I think patterns like `specialize (H x y z) as name` or `specialize (H x y z); intro name` don't signal prover's intention very well, so I tend to write `assert (name := H x y z)` instead.
- `specialize (H x y z) as [a b c]` doesn't signal intention as well as `destruct (H x y z) as [a b c]`.
- In a few places, `specialize` is used together with `rewrite` to achieve what can be better done with `replace x with y in H by tac`.
- Probably many more examples of this kind that I don't remember.

In general, `specialize` is a good and useful tactic, but your use of it (I know it was you!) is often suboptimal. The fact that I was able to get rid of around 700 lines by refactoring around the places where `specialize` occurs is a good evidence of this, I think.

palmskog (commented):
The proof assistant community has focused on two properties of proof scripts that improve maintainability: structure and predictability. Structure is mostly about hierarchical organization of code with indenting, bullets, and special keywords like `by` (SSReflect and Isabelle's Isar are the main examples here). Predictability is mostly about a proof script failing quickly once a definition is changed, which can be accomplished by tacticals like `by` and subproofs like `{ ... }`. The main example of predictability is the MathComp library, which has needed very modest proof script changes across 10+ major Coq versions and restructuring of definitions.

I don't see how this PR improves structure or predictability. While the number of proof script lines can matter at large scale, a lot of the removals here are of empty lines, so the fewer number of actual code lines overall seem modest. Overall, this means that I'm not prepared to approve the PR.

wkolowski:
It seems this PR is dead and not well-liked.

PR #106, opened 2022-09-09
Equations-stuff

traiansf:
* Added `inspect` and `eq:` notation for `Equations`
* Extracted `NatExtras` from the dev repo

### DIFF
--- a/theories/VLSM/Lib/EquationsExtras.v
+++ b/theories/VLSM/Lib/EquationsExtras.v
@@ -1,0 +1,23 @@
+From Equations Require Export Equations.
+
+(* Definition of [inspect] available in Equations as of version 1.3+8.16.
+  Notation <<eq:>> not yet available (exists only in examples).
+*)
+
+(**
+  [inspect x] allows to pattern-match x while retaining a propositional equality with [x].
+
+  See https://github.com/mattam82/Coq-Equations/blob/v1.3-8.16/theories/Prop/Logic.v#L31
+*)
+Definition inspect {A} (a : A) : {b | a = b} :=
+  exist _ a eq_refl.
+
+(**
+  The [inspect] definition is used to pack a value with a proof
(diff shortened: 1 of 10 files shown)


theories/VLSM/Lib/EquationsExtras.v
```diff
@@ -0,0 +1,23 @@
+From Equations Require Export Equations.
```
palmskog, line 1: This is fine, but maybe we should also document in the coding conventions that *anyone* who is using Equations in the project should use: `From VLSM.Lib Require Import EquationsExtras.` rather than importing Equations directly?

palmskog (approved):
Looks good.

palmskog:
Ah, just want to flag up one issue here. Once we merge this PR, the Papillon build will cease to work.

theories/VLSM/Lib/EquationsExtras.v
wkolowski: If it's mandatory to always import `EquationsExtras`, then it probably should be mentioned in the coding conventions. But should it be mandatory?

wkolowski: If it's mandatory to always import `EquationsExtras`, then it probably should be mentioned in the coding conventions. But should it be mandatory?

palmskog: I don't see why it shouldn't be mandatory, since otherwise we get two different idioms to use Equations, with different facilities available. This makes it harder to use automatic tools and similar to keep things updated.

traiansf: IN my view this is only a temporary measure; the inspect definition is already part of the 1.3+8.16 release and they seemed to be in favor to inlcude an `eq:`-like notation in a future release as well.  When that happens, we can simply replace the include to `EquationExtras` with the direct include to `Equations`.

PR #113, opened 2022-09-28
An alternative approach to finite sets based on Higher Inductive Types

wkolowski:
Instead of using `list`s that satisfy `NoDup` or classes from `stdpp`, we can define finite sets as a very concrete inductive type akin to `list`, provided that we believe in HoTT (or just that quotient types are ok).

I think this could also help with the other kind of finite sets that we have issues with, namely the finite type of `index`es in UMO/MO/ELMO composition.

Please take a look and tell me what your feelings are about this.

Note: this file is self-contained, so I used `easy` and `now` from the standard library instead of stdpp's `done` and `by`.

Note 2: the only thing I forgot to mention is that I chose not to truncate `FinSet` (i.e. I didn't force it to satisfy UIP), but that shouldn't be a big issue for us since we care mostly about the practical side of it.

### DIFF
--- a/_CoqProject
+++ b/_CoqProject
@@ -24,6 +24,7 @@ theories/VLSM/Lib/FinSetExtras.v
 theories/VLSM/Lib/StdppListFinSet.v
 theories/VLSM/Lib/EquationsExtras.v
 theories/VLSM/Lib/NatExtras.v
+theories/VLSM/Lib/AlternativeFinSet.v
 
 theories/VLSM/Core/VLSM.v
 theories/VLSM/Core/Composition.v
--- a/theories/VLSM/Lib/AlternativeFinSet.v
+++ b/theories/VLSM/Lib/AlternativeFinSet.v
@@ -1,0 +1,304 @@
+(**
+  Some tools from Homotopy Type Theory. They are already defined in the standard
+  library under different names, but we use the HoTT names for simplicity.
+*)
+
+(** Transport is basically rewriting. *)
+Definition transport {A : Type} (P : A -> Type) {x y : A} (p : x = y) : P x -> P y :=
+match p with
+| eq_refl _ => fun u : P x => u
+end.
+
+(** Functions preserve equality. *)
+Definition ap
+  {A B : Type} (f : A -> B) {x y : A} (p : x = y) : f x = f y :=
+match p with
+| eq_refl => eq_refl
+end.
+
+(** Dependent functions preserve equality *)
+Definition apd
+  {A : Type} {P : A -> Type} (f : forall x : A, P x) {x y : A} (p : x = y)
+  : transport P p (f x) = (f y) :=
+match p with
+| eq_refl => eq_refl
+end.
+
+(** We define finite sets as a Private Inductive Type hidden inside a module. *)
+Module FinSet.
+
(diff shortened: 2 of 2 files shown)


bmmoore (commented):
It's an interesting approach, but I think this way of implementing quotients in Coq is not very good, because needing so many `Axiom` is not good for auditability, even though it seems to be sound as used here.If Coq gets better support for higher inductive types it would be interesting to revisit.

I see how the postulated paths collapse different representations of the same finite set, but for faithfulness how do you prove that it didn't accidentally connect distinct sets?

Perhaps we could get a slightly less elegant but more obviously sound representation by working with a setoid equivalence instead of working to force them to be `eq`. An additional cost would be proving stuff like `Proper (eq ==> same_finite_set ==> iff)` for any operations over finite sets.

I wonder if we even need to declare types at all, it seems that even with this higher inductive type version, the burden is on the user of the recursion principle to prove their reduction operation respects idempotence and the swap.

But if you are going to do that work anyway, we could also just prove a theorem about the ordinary list fold that says the result of the ordinary fold of your operators is the same for any pair of lists that have the same finite set of elements.

palmskog:
Private inductives are considered an experimental feature of Coq by the core maintainers, similarly to `SProp`. Hence, it's not recommended for use in applications where trustworthiness is paramount. To my knowledge, soundness of private inductives (seen from the outside) have only been proven/sketched on pen-and-paper. I think this line of argument is sufficient for the case against these kinds of sets, but another line is that we wouldn't leverage the typeclass inference machinery for handling sets, a loss in automation potential.

wkolowski:
@bmmoore

> If Coq gets better support for higher inductive types it would be interesting to revisit.

When that happens, all problems with finite sets will be instantly gone :)

> I see how the postulated paths collapse different representations of the same finite set, but for faithfulness how do you prove that it didn't accidentally connect distinct sets?

Hmm, I think this question is somewhat confused. These `Axiom`s ("path constructors" in HoTT parlance) define what it means for two sets to be equal. They cannot connect distinct sets, because whatever they connect is equal by definition.

Maybe you wonder whether `FinSet` is actually a good representation of finite sets? I think this is quite easy to see: finite sets of elements of type `A` are the the same thing as free commutative idempotent monoids with generators taken from `A`.

I took inspiration from this paper: https://vikraman.org/files/cmon22.pdf

> Perhaps we could get a slightly less elegant but more obviously sound representation by working with a setoid equivalence instead of working to force them to be `eq`. An additional cost would be proving stuff like `Proper (eq ==> same_finite_set ==> iff)` for any operations over finite sets.

Yes, of course you can work directly with setoids, but then you end up in the "setoid hell" where you have to manually manage all these `Proper` proofs.

> I wonder if we even need to declare types at all, it seems that even with this higher inductive type version, the burden is on the user of the recursion principle to prove their reduction operation respects idempotence and the swap.

Yeah, the burden of course is on the user, just as with the setoid approach. The benefit of the higher inductive approach is precisely that the management of these `Proper` proofs gets handled automatically.

@palmskog

> To my knowledge, soundness of private inductives (seen from the outside) have only been proven/sketched on pen-and-paper.

Can you provide a reference?

The feedback is more or less what I expected - it's a nice approach, but it's on the cutting edge of things and pretty suspicious as far as being obviously correct. I guess I'm going to close this PR for now.

PR #111, opened 2022-09-22
Adhering to comments coding conventions (partially)

DafinaTrufas:

PR #114, opened 2022-10-03
some manual fixes to coqdoc formatting and text

palmskog:
This is a followup to #111, to fix the other problems reported in that PR, and some other ones that showed up in the same review.


wkolowski (approved):
Looks good.

PR #124, opened 2022-10-17
Tiny change to avoid breaking the axiom diagnostics script

wkolowski:
Probably it would be better to make the axiom diagnostics script more robust, but that would take much longer :)


traiansf (approved):

bmmoore:
Also, the grammar is better this way.

PR #125, opened 2022-10-17
Standardize section names

wkolowski:
Now all sections are written like `sec_camel_case`.

### DIFF
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -59,6 +59,8 @@ Example:
 Section sec_step_relations.
 ```
 
+The role of the "sec_" prefix is to prevent name collisions for crossreferences in coqdoc documentation. It should not appear in names of non-sections.
+
 ### Type classes
 
 - CamelCase name
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -52,11 +52,11 @@ From Equations Require Import Equations.
 
 ### Sections
 
-- C-style name
+- C-style name prefixed with "sec_"
 
 Example:
 ```coq
-Section step_relations.
+Section sec_step_relations.
 ```
 
 ### Type classes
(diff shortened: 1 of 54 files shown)


traiansf (commented):
I don't have anything against it. but I remember Karl had some ideas about camel-casing section, so I'll let him approve.

palmskog (approved):
This looks reasonable to me, but please add some quick motivation in the coding conventions, e.g.,
> the `sec` prefix prevents name collisions for crossreferences in coqdoc documentation

bmmoore (changes requested):
It seems a bit inconsistent whether uppercase letters are used in section names that refer to CamelCase names or if the name is converted to snake_case. In many case you seem to convert names, like `Section ByzantineTraces` -> `Section sec_byzantine_traces`, but others are retained.
In `VLSM.Core.MessageDependences` there is both a `sec_full_message_dependencies` where `FullMessageDependencies` is defined, but then also a `sec_CompositeHasBeenObserved_dec` where an `Instance CompositeHasBeenObserved_dec` is defined.

Is there supposed to be a pattern here that sometimes allows CamelCase fragments?
If so it should be explained in the conventions.
If not, then as long as this PR is touching all section names we should make them all follow the "C-style" in the coding conventions.
(I think "C-style" means snake_case with uppercase only allowed for acronyms like `VLSM` or `ELMO`)

wkolowski:
@bmmoore I think that CamelCase should be allowed to appear in section names if it refers to an identifier which is CamelCase. So I renamed `ByzantineTraces` to `sec_byzantine_traces`, because I think it didn't refer to an identifier, but I kept `sec_CompositeHasBeenReceived_dec` - this should be interpreted as a "section about `CompositeHasBeenReceived_dec`". The `FullMessageDependencies` one looks like a mistake, I will correct it.

I can add a sentence of explanation, but a full explanation would result in some kind of occult identifier hermeneutics...

wkolowski (commented):
Self-review.

bmmoore:
Looks good. I found no other places where you changed a CamelCase part of a section name that was actually matching an identifier.

bmmoore (approved):
Approve

PR #126, opened 2022-10-18
Make comments more conformant to the coding conventions

wkolowski:


traiansf (approved):
Looks OK to me, although I'm not so sure about breaking lines after opening a regular multiline comment.

PR #128, opened 2022-10-24
Replace the finishers easy and now with done and by.

wkolowski:
I saw `easy` and `now` from the standard library used in a few places whereas we prefer `done` and `by` from stdpp.


palmskog (approved):
Fine by me.

PR #129, opened 2022-10-24
Standardize indentation for (co)inductive types

wkolowski:
The coding conventions said that we should have 1 space of indentation for constructors of inductives, but most code (and all new code) used zero spaces, whereas some old code used two spaces. I chose to follow the new code and use zero spaces everywhere and update the coding conventions accordingly.

In case you don't like my choice this can be changed to whatever other convention, I just wanted to make things consistent.

### DIFF
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -96,13 +96,14 @@ Program Instance base_params (p : param) : BaseParams := {
 }.
 ```
 
-### Inductive and CoInductive types
+### Inductive and coinductive types
 
 - C-style type name
 - CamelCase constructors
-- to avoid name clashes, constructor names can be prefixed with an abbreviation of the type name
+- to avoid name clashes, constructor names can be prefixed with an abbreviation of the type name followed by `_`
 - no indentation in constructor declarations
 - the first `|` in constructor declarations should not be omitted
+- it's preferred to include the sort annotation
 
 Example:
 ```coq
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -103,7 +103,7 @@ Program Instance base_params (p : param) : BaseParams := {
 - to avoid name clashes, constructor names can be prefixed with an abbreviation of the type name followed by `_`
 - no indentation in constructor declarations
 - the first `|` in constructor declarations should not be omitted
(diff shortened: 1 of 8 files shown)


bmmoore (approved):
I like this style. Changes look correct.

We don't have to standardize it, but do you also want to suggest a specific format (like `ABR_name`) for the type-abbreviation prefix on constructor names?

wkolowski:
@bmmoore

I don't think this is that important and so can be left out. I think I indeed use something like `ABC_name`, but while working on these changes I think I saw a lot of cases (mostly by Traian, I guess) where the abbreviations were all lowercase. The only commonality is `_` as the separator between the abbreviation and the name proper.

bmmoore:
I also don't think it's too important to specify. But having existing code use several different styles is exactly what you might dislike and want to fix with another "Standardize ..." PR.

CODING_CONVENTIONS.md
```diff
@@ -96,18 +96,20 @@ Program Instance base_params (p : param) : BaseParams := {
 }.
 ```
 
-### Inductive types
+### Inductive and CoInductive types
```
palmskog, line 99:
```suggestion
### Inductive and Coinductive types
```
Even though they are written `CoInductive` inside Coq, it's not syntactically correct in English...

palmskog (approved):
Fine by me.

CODING_CONVENTIONS.md
bmmoore: I think the Coq keywords are slightly more correct here, because we are specifically talking about indentation for the `Inductive` and `CoInductive` Vernacular forms, but maybe mark them up with backtics. The general notion of "inductive types" conceivably includes Mendler-style encodings like
```
Definition myList (A:Type) :=
  forall (X:Type), (option (A*X) -> X) -> X.
```

bmmoore: Oh, and I was also about to suggest "coinductive" should be in lower case as @wkolowski changed it to, but realized this is a section title, so it's allowed to capitalize it.

wkolowski: @bmmoore I don't consider your `myList` an inductive type and I think it shouldn't be considered as such, because you can't prove the induction principle for it, right? The word "encoding" appears there for a reason...

PR #130, opened 2022-10-25
Standardize records

wkolowski:
Indentation and style for `Record`s was pretty inconsistent across the codebase and often ignoring the coding conventions. I think in most new code we use the style with braces on separate lines, fields on separate lines with 2 spaces on indentation and the sort annotation not omitted, so I made everything consistent with that and updated the coding conventions to reflect it.


bmmoore (approved):
Looks good. (I didn't check whether other definitions should have been updated).

bmmoore:
Oops, actually you should include "and the sort annotation not omitted" in the CODING_CONVENTIONS.md

bmmoore (approved):
That does it.

CODING_CONVENTIONS.md
```diff
@@ -140,11 +140,14 @@ Lemma sync_some (s : vstate X) (from to : index) :
 
 - C-style name
 - CamelCase for constructor name
-- field declaration with C-style naming on separate line
+- field declaration with C-style naming on separate line, with 2 spaces of indentation
+- the `;` in the last field should not be omitted
+- the sort annotation should not be omitted
```
palmskog, line 145: I think this should not be a hard rule. The sort annotation can easily be misleading, since for example this declaration still leads to a type in `Set`:
```coq
Inductive color : Type = Red | Green.
```

theories/VLSM/Core/VLSMProjections/VLSMTotalProjection.v
```diff
@@ -147,12 +147,15 @@ Record VLSM_projection_type
...
-        state_project (finite_trace_last sX trX) = finite_trace_last (state_project sX) (trace_project trX)
-  }.
+  : Prop :=
+{
+  final_state_project :
+    forall sX trX,
+      finite_valid_trace_from X sX trX ->
+        state_project (finite_trace_last sX trX)
+          =
+        finite_trace_last (state_project sX) (trace_project trX);
```
palmskog, line 157: Not sure I understand why a whole line for the `=` is needed here. Why not just put it at the end of the previous line?

palmskog (approved):
Looks reasonable, but please consider comments.

theories/VLSM/Core/VLSMProjections/VLSMTotalProjection.v
wkolowski: I can rewrite it, but I think it looks better this way.

CODING_CONVENTIONS.md
palmskog: I think it's better to write that: including the sort is *recommended*, in particular for `Prop`. "Should" is basically a synonym for "must" in many English specifications...

wkolowski: Ok, I will include the "recommended, particularly for `Prop`" version when I'll be merging this, and similarly for the other PRs.

PR #131, opened 2022-10-25
Standardize classes

wkolowski:
This is basically the same thing as in https://github.com/runtimeverification/vlsm/pull/130: the style of `Class`es was pretty inconsistent, so I attempted to make it more uniform, changing older code to match newer code.

### DIFF
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -70,7 +70,7 @@ Section sec_ELMOComponent_lemmas.
 - CamelCase name
 - field declaration with C-style naming on separate line, with 2 spaces of indentation
 - the `;` in the last field should not be omitted
-- the sort annotation should not be omitted
+- it's recommended to include the sort annotation, especially when it's `Prop`
 
 Example:
 ```coq
@@ -146,7 +146,6 @@ Lemma sync_some (s : vstate X) (from to : index) :
 - CamelCase for constructor name
 - field declaration with C-style naming on separate line, with 2 spaces of indentation
 - the `;` in the last field should not be omitted
-- the sort annotation should not be omitted
 - it's recommended to include the sort annotation, especially when it's `Prop`
 
 Example:
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -70,6 +70,7 @@ Section sec_ELMOComponent_lemmas.
 - CamelCase name
 - field declaration with C-style naming on separate line, with 2 spaces of indentation
 - the `;` in the last field should not be omitted
+- the sort annotation should not be omitted
 
 Example:
 ```coq
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -68,13 +68,15 @@ Section sec_ELMOComponent_lemmas.
 ### Type classes
 
 - CamelCase name
-- field declaration with C-style naming on separate line
+- field declaration with C-style naming on separate line, with 2 spaces of indentation
+- the `;` in the last field should not be omitted
 
 Example:
 ```coq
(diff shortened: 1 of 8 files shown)


theories/VLSM/Core/VLSM.v
```diff
@@ -38,15 +39,16 @@ Class VLSMType (message : Type) :=
...
-  ; s0 : Inhabited initial_state
-  ; initial_message_prop : message -> Prop
-  ; initial_message := { m : message | initial_message_prop m }
-  ; transition : label -> state * option message -> state * option message
-  ; valid : label -> state * option message -> Prop
-  }.
+Class VLSMMachine {message : Type} (vtype : VLSMType message) : Type :=
+{
+  initial_state_prop : state -> Prop;
+  initial_state := { s : state | initial_state_prop s };
```
bmmoore, line 45: Should this have a type annotation?

bmmoore (approved):
Looks fine if you don't want to require type annotations on all fields, but I think it should probably be consistent with the Record style.

theories/VLSM/Core/VLSM.v
wkolowski: Why not!

wkolowski:
@bmmoore What is the inconsistency?

bmmoore:
The consistency was having that line "the sort annotation should not be omitted" in the policy for one and not the other.

Also, I understood that line to require type annotations also on members that were defined with `:=`, which wasn't followed where you changed it. But maybe you just meant annotating the type of the record/class itself as one of `Prop`/`Set`/`Type`?

bmmoore (approved):
Looks good.

wkolowski:
@bmmoore Yes, by "sort annotation" I meant the one for the whole `Class`. For fields I wouldn't call these "sort" annotations but rather "type" annotations. But anyway I think this should be left for another PR, in which we'll decide when annotations should appear not only in the context of classes.

CODING_CONVENTIONS.md
```diff
@@ -68,13 +68,16 @@ Section sec_ELMOComponent_lemmas.
 ### Type classes
 
 - CamelCase name
-- field declaration with C-style naming on separate line
+- field declaration with C-style naming on separate line, with 2 spaces of indentation
+- the `;` in the last field should not be omitted
+- the sort annotation should not be omitted
```
palmskog, line 73: As for records in general, I think we should be flexible on the sort annotation, since it may not even make a difference. The main interest we have is when something is `Prop` and non-`Prop`, maybe you can add something to this effect?

theories/VLSM/Core/Equivocation.v
```diff
@@ -1380,7 +1377,7 @@ Class ComputableSentMessages `(vlsm : VLSM message) :=
 
 Global Hint Mode ComputableSentMessages - ! : typeclass_instances.
 
-Class ComputableReceivedMessages `(vlsm : VLSM message) :=
+Class ComputableReceivedMessages `(vlsm : VLSM message) : Type :=
```
palmskog, line 1380: Similar to the above classes, I don't think much is gained here by adding this `Type`. Would be better to just have: "if `Prop`, then say `Prop`, otherwise say nothing"

palmskog (approved):
Looks reasonable, but the added value for non-Prop is small.

theories/VLSM/Core/Equivocation.v
wkolowski: I'm strongly used to relying on having the sort annotation. If it's not there I start wondering "what's the sort of this" and it usually takes me a few seconds... much less efficient than reading the sort from the annotation.

CODING_CONVENTIONS.md
palmskog: As in the other PR, I think "recommended" is better than "preferred".

PR #132, opened 2022-10-27
Standardize the use of finishers across the VLSM codebase

wkolowski:
I made sure that all subgoals end with the use of `by`, refactoring some proofs as I went along.

Also with @bmmoore we agreed that we should write `by congruence`, `by itauto` even though `congruence` and `itauto` are succeed-or-fail style tactics.

Similarly we should also write `by firstorder`.

### DIFF
--- a/theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
@@ -166,11 +166,11 @@ Proof.
   - intros Hs sub_i.
     destruct Hs as [sX [HeqsX Hinitial]].
     subst.
-    apply Hinitial.
+  - by apply Hinitial.
   - intros Hs.
     exists (lift_sub_state fixed_byzantine_IM non_byzantine s).
     split.
-    + apply composite_state_sub_projection_lift_to.
+    + by apply composite_state_sub_projection_lift_to.
     + by apply (lift_sub_state_initial fixed_byzantine_IM non_byzantine).
 Qed.
 
@@ -199,12 +199,12 @@ Proof.
   apply induced_sub_projection_friendliness.
   apply induced_sub_projection_lift.
   intros s1 s2 Heq l om.
-  destruct om as [m|]; [|itauto].
+  destruct om as [m |]; [| by itauto].
   cbn.
-  destruct (sender m) as [v|]; [|itauto].
+  destruct (sender m) as [v |]; [| by itauto].
   simpl.
-  case_decide as HAv; [|itauto].
-  replace (s2 (A v)) with (s1 (A v)); [itauto|].
+  case_decide as HAv; [| by itauto].
+  replace (s2 (A v)) with (s1 (A v)); [by itauto |].
   exact (equal_f_dep Heq (dexist (A v) HAv)).
 Qed.
 
@@ -221,7 +221,7 @@ Lemma fixed_byzantine_trace_char1 is tr
 Proof.
   apply and_comm.
   apply (projection_friendly_trace_char (induced_sub_projection_is_projection _ _ _)).
-  apply fixed_non_byzantine_projection_friendliness.
+  by apply fixed_non_byzantine_projection_friendliness.
 Qed.
 
 End sec_fixed_byzantine_traces_as_projections.
@@ -273,7 +273,7 @@ Qed.
 Lemma non_byzantine_nodes_same_sym
   : forall sub_i, sub_IM IM non_byzantine sub_i = sub_IM fixed_byzantine_IM non_byzantine sub_i.
 Proof.
-  intro. symmetry. apply non_byzantine_nodes_same.
+  by intro; symmetry; apply non_byzantine_nodes_same.
 Qed.
 
 Lemma pre_loaded_fixed_non_byzantine_vlsm_full_projection
@@ -287,8 +287,8 @@ Proof.
   intros s1 Hs1 l1 om [Hom _].
   split; [| done].
   destruct om as [m |]; [| done].
-  destruct Hom as [Hsent | Hseeded]; [left | by right].
-  by eapply same_IM_composite_has_been_sent_preservation.
+  destruct Hom as [Hsent | Hseeded]; [| by right].
+  by left; eapply same_IM_composite_has_been_sent_preservation.
 Qed.
 
 Lemma pre_loaded_fixed_non_byzantine_vlsm_full_projection'
@@ -302,8 +302,8 @@ Proof.
   intros s1 Hs1 l1 om [Hom _].
   split; [| done].
   destruct om as [m |]; [| done].
-  destruct Hom as [Hsent | Hseeded]; [left | by right].
-  by eapply same_IM_composite_has_been_sent_preservation.
+  destruct Hom as [Hsent | Hseeded]; [| by right].
+  by left; eapply same_IM_composite_has_been_sent_preservation.
 Qed.
 
 (**
@@ -382,7 +382,7 @@ Proof.
             A sender fixed_byzantine_IM_sender_safety)
     ; [| done | done].
     eapply (VLSM_incl_valid_state); [| done].
-    eapply composite_pre_loaded_vlsm_incl_pre_loaded_with_all_messages.
+    by eapply composite_pre_loaded_vlsm_incl_pre_loaded_with_all_messages.
   - contradict HAv; clear -Hseeded Hsender.
     destruct Hseeded as [(i & Hi & Hm) _].
     unfold channel_authenticated_message in Hm.
@@ -434,7 +434,7 @@ Proof.
     simpl.
     rewrite @decide_True by done.
     cbn. unfold all_messages_transition.
-    inversion 1; itauto.
+    by inversion 1; itauto.
 Qed.
 
 (**
@@ -466,19 +466,18 @@ Proof.
     exists (lift_sub_label fixed_byzantine_IM non_byzantine l).
     exists (lift_sub_state fixed_byzantine_IM non_byzantine s).
     split.
-    + apply composite_label_sub_projection_option_lift.
-    + apply composite_state_sub_projection_lift.
+    + by apply composite_label_sub_projection_option_lift.
+    + by apply composite_state_sub_projection_lift.
     + by apply (VLSM_full_projection_input_valid pre_loaded_fixed_non_byzantine_vlsm_lift).
-  - intros l s om s' om' [_ Ht].
-    by apply induced_sub_projection_transition_preservation.
+  - by intros l s om s' om' [_ Ht]; apply induced_sub_projection_transition_preservation.
 Qed.
 
 Lemma fixed_non_byzantine_pre_loaded_eq
   : VLSM_eq fixed_non_byzantine_projection pre_loaded_fixed_non_byzantine_vlsm'.
 Proof.
   apply VLSM_eq_incl_iff. split.
-  - apply fixed_non_byzantine_pre_loaded_incl.
-  - apply pre_loaded_fixed_non_byzantine_incl.
+  - by apply fixed_non_byzantine_pre_loaded_incl.
+  - by apply pre_loaded_fixed_non_byzantine_incl.
 Qed.
 
 End sec_fixed_byzantine_traces_as_pre_loaded.
@@ -567,24 +566,22 @@ Lemma fixed_non_equivocating_incl_fixed_non_byzantine
 Proof.
   apply basic_VLSM_incl.
   - intros s (sX & <- & Hinitial) sub_i.
-    apply Hinitial.
+  - by apply Hinitial.
   - intro; intros.
     apply (VLSM_incl_input_valid fixed_non_equivocating_incl_sub_non_equivocating)
       in Hv as (Hs & _ & Hv).
-    apply sub_IM_no_equivocation_preservation in Hv as Hnoeqv.
-    2-4: done.
+    apply sub_IM_no_equivocation_preservation in Hv as Hnoeqv; [| done..].
     destruct Hnoeqv as [Hsent | Hseeded].
     + by eapply preloaded_composite_sent_valid.
     + apply initial_message_is_valid.
-      by right; split; [|eapply induced_sub_projection_valid_projection; apply Hv].
+      by right; split; [| eapply induced_sub_projection_valid_projection, Hv].
   - intros l s om Hv.
     apply (VLSM_incl_input_valid fixed_non_equivocating_incl_sub_non_equivocating)
        in Hv as (_ & _ & Hv).
     split.
     + by eapply induced_sub_projection_valid_preservation.
     + split; [| done].
-      apply sub_IM_no_equivocation_preservation in Hv as Hnoequiv.
-      2-4: done.
+      apply sub_IM_no_equivocation_preservation in Hv as Hnoequiv; [| done..].
       destruct om as [m |]; [| done].
       destruct Hnoequiv as [Hsent|Hseeded]; [by left | right].
       by split; [|eapply induced_sub_projection_valid_projection; apply Hv].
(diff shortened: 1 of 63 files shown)


bmmoore:
Reading the coding convention more carefully, I see it doesn't actually require `by congruence`. I was only complaining about using `itauto` without `by` because I mistakenly thought we were requiring `by` on `congruence` and nobody who instantly recognizes `itauto` as a finisher would be unfamiliar with `congruence`.

If we allow `congruence` as a finisher I don't necessarily mind the current state. But maybe if we hope to have readers unfamiliar with Coq maybe we should just completely standardize on `by` and `done`.

I can review this PR assuming the new policy.

But I would like @palmskog's (and @traiansf's if available) opinion on the style.

wkolowski:
@bmmoore Regarding finishers, I wanted to make sure that every proof and every subgoal is finished with `by` and that when we are using `; [ | ]` it is clear whether the given subgoal is solved (in which case `by` appears) or not. I changed to `by congruence` and `by itauto` everywhere, so please review with that in mind. In case we don't want it, it'll be easy to find-and-replace.

theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
```diff
@@ -566,31 +555,27 @@ Lemma fixed_non_equivocating_incl_fixed_non_byzantine
...
-      apply sub_IM_no_equivocation_preservation in Hv as Hnoequiv.
-      2-4: done.
+      apply sub_IM_no_equivocation_preservation in Hv as Hnoequiv; [| done..].
       destruct om as [m |]; [| done].
       destruct Hnoequiv as [Hsent|Hseeded]; [by left | right].
       by split; [|eapply induced_sub_projection_valid_projection; apply Hv].
   - intros l s om s' om' [_ Ht].
-    revert Ht.
-    by apply @induced_sub_projection_transition_preservation.
+    by revert Ht; apply @induced_sub_projection_transition_preservation.
```
palmskog, line 578: I don't really understand this refactoring - it doesn't make the proof easier to follow (in fact harder). Why wouldn't a single-line `revert` be allowed?

palmskog (commented):
I think there are two types of changes here, and I would prefer to see the former split out into a separate PR that can be merged quickly:
1. adding missing `by` finishers, e.g., for `firstorder` and so on
2. other changes that involve chaining tactics that were previously unchained and so on

I'm *not* on board with the second kind of changes. I can see no obvious gain from merging certain kinds of sentences. For example, having many tactics per line can in many cases just make proof failures harder to debug.

theories/VLSM/Core/ELMO/ELMO.v
```diff
@@ -332,10 +332,9 @@ Proof.
   - intros l s im s' om [(Hvsp & Hovmp & Hv) Ht] m; cbn in *.
     destruct l, im; cbn in *; [| by exfalso; inversion Hv..|];
     inversion_clear Ht; destruct s; cbn.
-    + rewrite decide_False; cbn; firstorder congruence.
+    + by rewrite decide_False; cbn; firstorder congruence.
     + rewrite decide_True by done; cbn.
-      unfold Message; rewrite elem_of_cons.
-      by firstorder congruence.
+      by unfold Message; rewrite elem_of_cons; firstorder congruence.
```
palmskog, line 337: Another example of a change that makes no sense to me, and I think doesn't have much to do with the topic of the PR.

theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
bmmoore: I think it makes more sense to consider `revert H; apply lem` as a unit - I understand it as "apply lem with H as some(or the penultimate) argument", and short for some `apply (lem _ ... _ H)`, or being a plain-tactics version the ssreflect `apply: lem H`.

But I agree this sort of change doesn't fit the overall aim of the PR, and that chaining more tactics isn't always an improvement.

wkolowski: @bmmoore is right - the only purpose of this `revert` is to make it possible to use the following `apply`, so they should be considered as a unit. And since the following `apply` finishes the proof, it gets touched in the PR, so I think it is in scope of this PR.

palmskog (changes requested):
@wkolowski after discussing with @traiansf I'd like to explicitly request a split PR here. Please restrict this PR to only doing the minimal changes to add finishers (without combining tactic sentences unless it's completely unavoidable), and do a different PR with other tactic changes.

theories/VLSM/Core/ELMO/ELMO.v
bmmoore: I don't see how that affects what should be joined with `;` or `.`. In that sense I would say everything after the `+` is instructions for finishing the proof, and it wouldn't work to say any bullet without sub-bullets should all be chained together.
I think the major questions are what makes sense to think of as conceptually single or similar steps, and where a reader might want to pause and view the proof state.

Any sequence of `unfold`, `rewrite`, and simplification tactics is usually a reasonable unit because it always produces a goal that should be equivalent to the original. And it's a natural place to step back and forth and check if you follow the equivalence or need to look up definitions - which is much easier if you can actually look at the term afterwards, instead of having to imagine it.

While `firstorder congruence` means "it may not be immediately obvious, but this is a (first-order+equality) tautotology". So it's not the same sort of thing.

theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
bmmoore: I'm defending it as good refactoring, but I don't think it's necessary for fixing nonstandard usage of `by` if we're splitting the PR now.

I see nothing incorrect about the usage of `by` with the `apply` in the old code. It's certainly a line you would have to inspect to check if finishers were used correctly everywhere, but I don't think it needed to be changed.

theories/VLSM/Core/ELMO/ELMO.v
wkolowski: My current philosophy is that each line should only contain a single atomic logical step, which can be composed out of many tactics with `;`. Bureaucracy and finishers are not atomic logical steps, so they should be joined to the really important part with `;`. So as I said, in this example the core is the `rewrite`, the bureaucracy is the `unfold` and `by` and `firstorder congruence` are the finisher, so I think it makes sense to chain them.

wkolowski:
I will try to filter this PR to only contain the changes you are asking for. Recently I learned about `git add --patch` - let's see how much it can do here.

bmmoore (commented):
The changes I've checked seem to okay either as fixing a finisher, or an acceptable refactoring that doesn't fix a finisher, but if those categories will become separate PRs I will wait until it is split.

wkolowski:
@palmskog @bmmoore Please review and tell me if it's better now. I tried to remove the changes that Karl didn't like from this PR using `git add --patch`, but some may have been overlooked. Also I'm afraid that the other refactorings might be gone, as I forgot to save them somewhere before I did `git push --force`.

theories/VLSM/Core/Equivocation.v
```diff
@@ -782,9 +780,7 @@ Proof.
   intro m.
   destruct Htr as [Htr Hinit].
   rewrite (oracle_partial_trace_update oracle_props _ _ _ Htr).
-  assert (~oracle s0 m).
-  apply oracle_props, Hinit.
-  itauto.
+  by firstorder.
```
palmskog, line 783: `firstorder` is not really considered a good tactic anymore (it's barely maintained), so I disagree that this is an improvement. However, it's a very minor occurrence in the grand scheme of this PR.

theories/VLSM/Core/Equivocation/WitnessedEquivocation.v
```diff
@@ -526,18 +525,15 @@ Proof.
       destruct Htr'_item as [Htr'_item Hinit].
       apply finite_valid_trace_from_to_app_split in Htr'_item.
       destruct Htr'_item as [Htr' Hitem].
-      spec IHn.
-      { split; [| done].
-        by apply finite_valid_trace_from_to_forget_last in Htr'.
-      }
+      spec IHn; [by split; [| done]; apply finite_valid_trace_from_to_forget_last in Htr' |].
```
palmskog, line 528: I disagree that this is an improvement. The previous formulation is much easier for debugging failing proofs.

theories/VLSM/Core/Equivocation/WitnessedEquivocation.v
```diff
@@ -754,14 +749,10 @@ Proof.
...
         (tr ++ [{| l := l; input := iom; destination := sf; output := oom |}])
         []).
-    spec Heqv; [apply app_nil_r|].
-    destruct iom as [im|].
-    2:{
-      repeat split
-      ; [done | apply option_valid_message_None | done..].
-    }
+    spec Heqv; [by apply app_nil_r |].
+    destruct iom as [im|]; [| by repeat split; [| apply option_valid_message_None | ..]].
```
palmskog, line 753: This has now become a very deeply nested goal structure. Would be better to actually introduce bullets so we can keep the separate goal from previously.

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/FixedEquivocation.v
```diff
@@ -1139,10 +1144,7 @@ Proof.
       _ n item
     ) as Hitem_equivocating.
   clear Hdescriptors n.
-  spec Hitem_equivocating.
-  { subst.
-    rewrite !elem_of_app.
-    right. left. left. }
+  spec Hitem_equivocating; [by subst; rewrite !elem_of_app; right; left; left |].
```
palmskog, line 1147: I disagree that this is an improvement, again for reason of more difficult proof debugging.

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/FixedEquivocationSimulation.v
```diff
@@ -441,11 +441,7 @@ Proof.
         ).
   - clear isX sX trX HtrX. intro; intros.
     specialize (equivocators_fixed_equivocations_vlsm_incl_PreFree IM []) as HinclE.
-    destruct iom as [im|].
-    2: {
-      exists []. exists eqv_state_s.
-      split; [constructor |]; eauto.
-    }
+    destruct iom as [im |]; [| by exists [], eqv_state_s; split; [constructor |]; eauto].
```
palmskog, line 444: Like a similar comment elsewhere, this is not an improvement to me.

theories/VLSM/Core/Equivocators/Composition/SimulatingFree/SimulatingFree.v
```diff
@@ -317,11 +315,7 @@ Proof.
     by elim (no_initial_messages_in_IM i mi).
   - clear isX sX trX HtrX.
     intro; intros.
-    destruct iom as [im|].
-    2: {
-      exists [], eqv_state_s.
-      by split; [constructor |].
-    }
+    destruct iom as [im |]; [| by exists [], eqv_state_s; split; [constructor |]].
```
palmskog, line 318: Same non-improvement pattern as elsewhere.

theories/VLSM/Core/VLSM.v
```diff
@@ -1628,12 +1622,12 @@ Proof.
...
   clear Hinit.
   induction Htl.
-  - constructor. by apply initial_state_is_valid.
+  - by constructor; apply initial_state_is_valid.
   - apply finite_valid_trace_from_to_app with s; [done |].
     apply finite_valid_trace_from_to_singleton.
     apply finite_valid_trace_init_to_emit_valid_state_message in Htl1.
     apply finite_valid_trace_init_to_emit_valid_state_message in Htl2.
-    repeat split; try red; eauto.
+    by firstorder.
```
palmskog, line 1630: As elsewhere, I don't think this is an improvement. Can't `itauto` be a replacement somehow?

theories/VLSM/Lib/StreamExtras.v
```diff
@@ -731,8 +716,5 @@ Proof.
   exists n.
   apply ForAll1_forall.
   rewrite ForAll1_forall in Hn.
-  intros n' Hn'.
-  elim (Hn n').
-  revert Hn'.
-  apply HPQ.
+  by firstorder.
```
palmskog, line 719: As usual, I don't think this is a gain. Changing to use `itauto` would be.

palmskog (commented):
This looks a whole lot better now, but please address my remaining comments in some way. If that's done, I'm prepared to approve this.

theories/VLSM/Core/Equivocation.v
wkolowski: Maybe leave getting rid of `by firstorder` for a separate PR? I checked and `itauto` doesn't work here.

theories/VLSM/Core/Equivocation/WitnessedEquivocation.v
wkolowski: Ok, I reverted it back. But what about

```suggestion
      spec IHn; [by split; [apply finite_valid_trace_from_to_forget_last in Htr' |] |].
```

Do you think it is just as bad?

theories/VLSM/Core/Equivocation/WitnessedEquivocation.v
wkolowski: What about

```suggestion
    destruct iom as [im |]; [| by repeat split; auto using option_valid_message_None].
```

It would be sad to introduce bullets, as this would indent to following ~50 lines.

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/FixedEquivocation.v
wkolowski: Do you think any improvements to the `by`-clause can salvage this?

```suggestion
    by subst; rewrite !elem_of_app, elem_of_cons; auto.
```
Or should I just rever to the previous version?

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/FixedEquivocationSimulation.v
wkolowski:
```suggestion
    destruct iom as [im |]; [| by exists [], eqv_state_s; split; constructor].
```

Any better?

theories/VLSM/Core/VLSM.v
wkolowski: `itauto` doesn't work here, probably because it can't do this `red` step that was there in the original proof that `firstorder` can handle somehow.

theories/VLSM/Lib/StreamExtras.v
wkolowski: `itauto` doesn't work here. I even tried to use `itauto` after every step of the manual proof and it still doesn't work. I think `itauto` can't do first-order reasoning.

wkolowski:
@palmskog I tried to address your suggestions. I changed some of the nested stuff to be less nested. Please tell me whether the changes are good enough or whether I should revert.

I also think we should deal with `by firstorder` in a separate PR, as there are many more occurrences of it besides this PR. Tell me whether this is ok or whether I should revert.

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/FixedEquivocationSimulation.v
```diff
@@ -441,24 +441,20 @@ Proof.
         ).
   - clear isX sX trX HtrX. intro; intros.
     specialize (equivocators_fixed_equivocations_vlsm_incl_PreFree IM []) as HinclE.
-    destruct iom as [im|].
-    2: {
+    destruct iom as [im |]; [| by exists [], eqv_state_s; split; constructor].
+    + destruct HcX as [Hsent | Hemitted]; [| done].
```
palmskog, line 445: I don't really understand what is going on here, there is a single `+` goal? Please take a look.

palmskog (approved):
There are several refactorings I'm not very fond of here (subgoals being replaced with one-liners with many changed tactics). But there seems to be reasonable improvement.

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/FixedEquivocationSimulation.v
wkolowski: Good catch, indeed there was a superfluous `+`.

PR #133, opened 2022-10-28
Wrap lines longer than ~100 characters

wkolowski:
The coding conventions say lines should be at most 80 characters. It would be nice to follow this, but it's too much work, so I decided to wrap only these lines that don't fit into my CoqIDE window, which is around 100 characters long.

### DIFF
--- a/theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
@@ -2,10 +2,12 @@ From stdpp Require Import prelude finite.
 From Coq Require Import FunctionalExtensionality Reals.
 From VLSM.Lib Require Import Preamble StdppListSet Measurable ListSetExtras RealsExtras.
 From VLSM.Core Require Import VLSM MessageDependencies VLSMProjections Composition ProjectionTraces.
-From VLSM.Core Require Import SubProjectionTraces AnnotatedVLSM ByzantineTraces.FixedSetByzantineTraces.
-From VLSM.Core Require Import Validator Equivocation Equivocation.FixedSetEquivocation.
+From VLSM.Core Require Import SubProjectionTraces AnnotatedVLSM Validator Equivocation.
+From VLSM.Core Require Import ByzantineTraces.FixedSetByzantineTraces.
+From VLSM.Core Require Import Equivocation.FixedSetEquivocation.
 From VLSM.Core Require Import Equivocation.LimitedMessageEquivocation.
-From VLSM.Core Require Import Equivocation.MsgDepLimitedEquivocation Equivocation.TraceWiseEquivocation.
+From VLSM.Core Require Import Equivocation.MsgDepLimitedEquivocation.
+From VLSM.Core Require Import Equivocation.TraceWiseEquivocation.
 
 (** * VLSM Compositions with Byzantine nodes of limited weight
 
@@ -57,7 +59,8 @@ Definition limited_byzantine_trace_prop
   exists byzantine, fixed_limited_byzantine_trace_prop s tr byzantine.
 
 Context
-  {is_equivocating_tracewise_no_has_been_sent_dec : RelDecision (is_equivocating_tracewise_no_has_been_sent IM (fun i => i) sender)}
+  {is_equivocating_tracewise_no_has_been_sent_dec :
+    RelDecision (is_equivocating_tracewise_no_has_been_sent IM (fun i => i) sender)}
   (limited_constraint := tracewise_limited_equivocation_constraint IM (Ci := Ci) sender)
   (Limited : VLSM message := composite_vlsm IM limited_constraint)
   (Hvalidator: forall i : index, component_message_validator_prop IM limited_constraint i)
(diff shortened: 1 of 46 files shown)


theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
```diff
@@ -2,10 +2,10 @@ From stdpp Require Import prelude finite.
...
 From VLSM.Lib Require Import Preamble StdppListSet Measurable ListSetExtras RealsExtras.
 From VLSM.Core Require Import VLSM MessageDependencies VLSMProjections Composition ProjectionTraces.
-From VLSM.Core Require Import SubProjectionTraces AnnotatedVLSM ByzantineTraces.FixedSetByzantineTraces.
-From VLSM.Core Require Import Validator Equivocation Equivocation.FixedSetEquivocation.
-From VLSM.Core Require Import Equivocation.LimitedMessageEquivocation.
-From VLSM.Core Require Import Equivocation.MsgDepLimitedEquivocation Equivocation.TraceWiseEquivocation.
+From VLSM.Core Require Import SubProjectionTraces AnnotatedVLSM Validator Equivocation.
+From VLSM.Core Require Import ByzantineTraces.FixedSetByzantineTraces.
+From VLSM.Core.Equivocation Require Import FixedSetEquivocation LimitedMessageEquivocation.
+From VLSM.Core.Equivocation Require Import MsgDepLimitedEquivocation TraceWiseEquivocation.
```
palmskog, line 8: The reason why we write `From VLSM.X` and `From VLSM.X.Y` is so we can use tools to modify/minimize the Requires and Imports. I don't think this should be changed here - instead we can have more lines with `From ...`.

theories/VLSM/Core/ByzantineTraces/LimitedByzantineTraces.v
```diff
@@ -238,9 +245,15 @@ Lemma validator_limited_non_byzantine_traces_are_limited_non_equivocating s tr
...
-      finite_trace_sub_projection IM selection_complement tr = finite_trace_sub_projection IM selection_complement btr.
+        (sum_weights (remove_dups selection) <= `threshold)%R
+          /\
+        composite_state_sub_projection IM selection_complement s
+          =
+        composite_state_sub_projection IM selection_complement bs
+          /\
+        finite_trace_sub_projection IM selection_complement tr
+          =
+        finite_trace_sub_projection IM selection_complement btr.
```
palmskog, line 256: I think this is much more difficult to read than the previous sentences. Why do we need standalone lines for all connectives?

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/LimitedEquivocationSimulation.v
```diff
@@ -1,15 +1,17 @@
...
+From VLSM.Lib Require Import StdppListSet RealsExtras Measurable.
 From VLSM.Core Require Import VLSM VLSMProjections Composition AnnotatedVLSM.
-From VLSM.Core Require Import Equivocation Equivocation.FixedSetEquivocation.
-From VLSM.Lib Require Import Measurable.
-From VLSM.Core Require Import Equivocation.TraceWiseEquivocation Equivocation.LimitedMessageEquivocation Equivocation.MsgDepLimitedEquivocation.
-From VLSM.Core Require Import MessageDependencies Equivocation.WitnessedEquivocation.
-From VLSM.Core Require Import Equivocators.Composition.EquivocatorsComposition Equivocators.Composition.EquivocatorsCompositionProjections.
+From VLSM.Core Require Import Equivocation MessageDependencies.
+From VLSM.Core.Equivocation Require Import TraceWiseEquivocation LimitedMessageEquivocation.
+From VLSM.Core.Equivocation Require Import
```
palmskog, line 8: See other comment, I think we should use additional lines here rather than changing to `From VLSM.X.Y`.

palmskog (commented):
This mostly looks reasonable, but see comments about `From VLSM.X.Y`.

palmskog (commented):
If this was actually about doing the minimal changes to wrap lines, I'd have no problem approving it. But this is basically a code reformatting PR with a line wrapping title (see comments). I advocate doing the minimal line wrapping here and splitting out code formatting to another PR.

palmskog (changes requested):
@wkolowski in the same vein as #132, please restrict this PR to only fixing line wrapping in the most straightforward way, and put the remaining code formatting changes in another PR.

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/FixedEquivocation.v
```diff
@@ -618,12 +632,18 @@ Proof.
     apply equivocators_fixed_equivocations_vlsm_incl_free.
   }
 
-  assert (HtrX_Pre : finite_valid_trace (pre_loaded_with_all_messages_vlsm Free) (equivocators_state_project IM initial_descriptors is) trX ).
-  { revert HtrX_Free. apply VLSM_incl_finite_valid_trace. apply vlsm_incl_pre_loaded_with_all_messages_vlsm. }
+  assert (HtrX_Pre : finite_valid_trace (pre_loaded_with_all_messages_vlsm Free)
+                      (equivocators_state_project IM initial_descriptors is) trX ).
```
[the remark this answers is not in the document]
wkolowski: The idea is to indent the second line two spaces more than the expression to which it is argument, ie. `finite_valid_trace ...`. But probably it makes more sense to indent just two spaces from the `assert`.

palmskog (approved):
OK, this now looks reasonable (i.e.,I think the title is now accurate, and the content provides a benefit).

PR #137, opened 2022-11-11
Renamed full projections to embeddings

traiansf:


palmskog (approved):
I can certify this is a very clean renaming.

PR #138, opened 2022-11-11
VLSM Stuttering Embeddings

traiansf:
Additions:
- non-empty lists
- stream_concat for a stream of non-empty lists
- VLSM Stuttering projections --- a new version of VLSM projections which should stand between partial_projections and total_projection and would allow a trace item to transform into a list of trace items. I think that VLSMTotalProjection can be restated in terms of the stuttering projection; might do that after this PR is merged.


theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,265 @@
...
+
+(** A straight-forward inductive definition of non-empty lists. *)
+Inductive ne_list (A : Type) : Type :=
+| ne_one : A -> ne_list A
+| ne_cons : A -> ne_list A -> ne_list A.
+
+Arguments ne_one {_} _ : assert.
+Arguments ne_cons {_} _ _ : assert.
+
+Global Infix ":::" := ne_cons (at level 60, right associativity).
```
palmskog, line 12:
```suggestion
#[export] Infix ":::" := ne_cons (at level 60, right associativity).
```
I think the usual convention is to export notations like this, not add them for everyone that recursively touches a module.

theories/VLSM/Lib/Preamble.v
```diff
@@ -211,6 +211,63 @@ Proof.
...
+  (bound : nat)
+  (Hbound : P bound)
+  : nat :=
+  (fix helper (dif : nat) :=
+        match dif with
+        | 0 => bound
+        | S dif' =>
+          if decide (P (bound - dif)) then bound - dif else helper dif'
+        end)
+      bound.
```
palmskog, line 230: For readability, how about writing this with a proper `let` instead? `let fix helper ... in helper bound`

palmskog (approved):
Looks reasonable, but please consider the comments.

theories/VLSM/Core/VLSMProjections/VLSMStutteringEmbedding.v
```diff
@@ -0,0 +1,726 @@
...
+  We say that a stuttering embedding is established from VLSM <<X>> to VLSM <<Y>>
+  (sharing the same messages) if there exists maps <<state_project>>
+  taking <<X>>-states to <<Y>>-states, and <<transition_item_project>>, taking
+  a transition from <<X>> to a list of transitions in <<Y>>, such that:
+
+  - [transition_item_project_consistency]: the state-translation of the destination
+    of a transition is the final state of the translation of the transition.
+
+  - [stuttering_embedding_preserves_valid_trace]s: traces obtained by
+    concatenating the translations of transition of a valid trace are valid.
```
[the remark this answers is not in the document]
traiansf: Thanks! this probably means that some other similar lists should be refactored (as I copied this from elsewhere).

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,265 @@
...
+  (* for some strange reason lia doesn't work directly, so... this hack: *)
+  assert (Hle : forall a b c, a >= 1 -> b >= c -> a + b >= S c) by lia.
+  by apply Hle; [apply ne_list_min_length |].
+Qed.
+
+(**
+  An alternative inductive definition of non-empty lists using a single
+  constructor.
+*)
+Inductive Nel (A : Type) : Type :=
```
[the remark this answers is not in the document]
traiansf: Hm... I already have a `NeList` so I fear naming this `NEList` would make it confusing.
Since I don't plan to use this much (I just added it because Wojciech mentioned it), I see two options:
- giving it a longer name, like: `NonEmptyList`
- removing it altogether.

Sugegstions?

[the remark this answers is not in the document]
palmskog: I vote for a longer name like `NonEmptyList`.

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,265 @@
...
+  (* for some strange reason lia doesn't work directly, so... this hack: *)
+  assert (Hle : forall a b c, a >= 1 -> b >= c -> a + b >= S c) by lia.
+  by apply Hle; [apply ne_list_min_length |].
+Qed.
+
+(**
+  An alternative inductive definition of non-empty lists using a single
+  constructor.
+*)
+Inductive NonEmptyList (A : Type) : Type :=
```
palmskog, line 224: One more comment, this is one of those types that without closer scrutiny could turn out to be empty. Maybe you want to prove that it's inhabited?

wkolowski: I disagree. It's quite obvious that it's inhabited, given that you can use `None` to fill the second argument of `nel_cons`.

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,268 @@
...
+  induction l using rev_ind; [done |].
+  rewrite mbind_app, list_filter_map_app, map_app, mjoin_app, IHl.
+  cbn; clear IHl.
+  f_equal; rewrite app_nil_r.
+  destruct (decide _); cbn; cycle 1.
+  - symmetry; destruct (decide (f x = [])); [done |].
+    by contradict n.
+  - subst f' P; unfold list_function_restriction, ne_list_from_non_empty_list; cbn in p.
+    remember (proj2_dsig _) as Hp; clear HeqHp; cbn in Hp.
+    unfold proj1_sig, dexist.
```
wkolowski, line 203:
```suggestion
    simpl.
```

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,268 @@
...
+Qed.
+
+Lemma ne_list_concat_min_length
+  {A : Type}
+  (l : list (ne_list A))
+  : length (mjoin (map ne_list_to_list l)) >= length l.
+Proof.
+  induction l; cbn; [by lia |].
+  rewrite app_length.
+  (* for some strange reason lia doesn't work directly, so... this hack: *)
```
wkolowski, line 215: The reason is not strange. The problem is that `lia` doesn't know that `length (ne_list_to_list a)` is `>= 1`.

You should use a lemma here instead of an `assert`. Maybe an "unroll" lemma that relates `length` and `ne_list_length`?

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,268 @@
...
+Arguments nel_cons {_} _ _ : assert.
+
+#[export] Instance NonEmptyList_inhabited `{Inhabited A} : Inhabited (NonEmptyList A) :=
+  populate (nel_cons inhabitant None).
+
+Definition nel_hd `(l : NonEmptyList A) : A :=
+  match l with (nel_cons a _) => a end.
+
+Definition nel_tl `(l : NonEmptyList A) : option (NonEmptyList A) :=
+  match l with (nel_cons _ l') => l' end.
```
wkolowski, line 236: Maybe you should define `NonEmptyList` as a record instead? I think the way you did it doesn't give you the induction principle anyway.

wkolowski (changes requested):
This is a review of everything besides `theories/VLSM/Core/VLSMProjections/VLSMStutteringEmbedding.v`.

I opened a PR with suggested changes. The changes I request are described in some of my comments.

theories/VLSM/Lib/StreamExtras.v
```diff
@@ -718,3 +743,64 @@ Proof.
   rewrite ForAll1_forall in Hn.
   by firstorder.
 Qed.
+
+Definition stream_prepend {A} (nel : ne_list A) (s : Stream A) : Stream A :=
+  (cofix prepend (l : ne_list A) :=
+    Cons (ne_list_hd l) (from_option prepend s (ne_list_tl l))) nel.
```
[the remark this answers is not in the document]
wkolowski: I did check a bit and I still don't understand why using `CoFixpoint` here would make the following definition of `stream_concat` non-productive. @palmskog Any ideas?

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,268 @@
...
+  | ne_cons a _ => a
+  end.
+
+Definition ne_list_tl {A} (l : ne_list A) : option (ne_list A) :=
+  match l with
+  | ne_one _ => None
+  | ne_cons _ tl => Some tl
+  end.
+
+Definition ne_list_foldr {A B} (f : A -> B -> B) (b : B) (l : ne_list A) :=
```
wkolowski, line 26: I forgot to mention this, but why aren't you using `Fixpoint` here?

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,268 @@
...
+  (fix fold (l : ne_list A) :=
+    match l with
+    | ne_one a => f a b
+    | ne_cons a nel => f a (fold nel)
+    end
+  ) l.
+
+Definition ne_list_foldl {A B} (b : B) (f : B -> A -> B) (l : ne_list A) :=
```
wkolowski, line 34: As above, why `Definition` with `fix` instead of `Fixpoint`?

theories/VLSM/Lib/Preamble.v
```diff
@@ -194,6 +194,63 @@ Proof.
...
+  `{forall n, Decision (P n)}
+  (bound : nat)
+  (Hbound : P bound)
+  : nat :=
+  let fix helper (dif : nat) :=
+        match dif with
+        | 0 => bound
+        | S dif' =>
+          if decide (P (bound - dif)) then bound - dif else helper dif'
+        end
```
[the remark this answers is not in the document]
palmskog: `let fix` is a standard technique see for example: https://github.com/coq-community/fourcolor/blob/master/theories/redpart.v#L1270-L1273

theories/VLSM/Lib/NeList.v
traiansf: People (see, e.g., stdpp) seem to like delegating things to a `fix` construct having only the iterating parameter as an argument. I don't have a clear preference towards any of the approaches.

theories/VLSM/Lib/Preamble.v
```diff
@@ -194,6 +194,63 @@ Proof.
   by split; [| lia].
 Qed.
 
+(**
+  Given a decidable property on naturals and a natural number on which the
+  property holds, we can compute the least natural number on which it holds.
+*)
+Definition compute_minimal_among_le
```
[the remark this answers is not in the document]
traiansf: I've named it `find_least_among`.  The `among` is kept to keep some resemblance with `minimal_among`

theories/VLSM/Lib/StreamExtras.v
```diff
@@ -718,3 +743,64 @@ Proof.
...
+  {A : Type}
+  (l : ne_list A)
+  (s : Stream A)
+  (n : nat)
+  (Hle : n <= ne_list_length l)
+  : stream_prefix (stream_prepend l s) n = list_prefix (ne_list_to_list l) n.
+Proof.
+  revert n Hle; induction l; intros [| n] Hle; cbn; [done | | done |].
+  - by cbn in Hle; replace n with 0 by lia.
+  - by cbn in *; rewrite IHl; [| cbv in *; lia].
```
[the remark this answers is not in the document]
traiansf: I usually don't use `cbv`. But it seems that sometimes `lia` needs something stronger than `cbn`.

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,268 @@
+From stdpp Require Import prelude.
+From VLSM.Lib Require Import ListExtras StdppExtras.
+
+(** A straight-forward inductive definition of non-empty lists. *)
+Inductive ne_list (A : Type) : Type :=
+| ne_one : A -> ne_list A
+| ne_cons : A -> ne_list A -> ne_list A.
```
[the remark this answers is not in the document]
traiansf: My habit so far was to prefix constructors (and field names) with an acronym of the type, to make them unique. I guess the acronym here would have better been `nel_` instead of `ne_`. I'll do that, and also change `ne_one` to `nel_singl`.

theories/VLSM/Lib/NeList.v
traiansf: Well, it start being stranger to me if instead of the code there I write:
  `specialize (ne_list_min_length a); unfold ne_list_length; lia.`
Which I would expect to finish the proof; instead lia complains about not being able to find a witness.

theories/VLSM/Lib/NeList.v
```diff
@@ -0,0 +1,268 @@
...
+  ne_list_foldr cons [] nel.
+
+Definition ne_list_length {A} (nel : ne_list A) := length (ne_list_to_list nel).
+
+Lemma ne_list_min_length {A} (nel : ne_list A) : ne_list_length nel >= 1.
+Proof. by destruct nel; cbn; lia. Qed.
+
+Lemma ne_list_to_list_unroll {A} (a : A) (nel : ne_list A) :
+  ne_list_to_list (a ::: nel) = a :: ne_list_to_list nel.
+Proof. done. Qed.
```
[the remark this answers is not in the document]
traiansf: maybe I don't, but currently it helps make the two proofs I used it in smoother (using sequence of rewrites).

theories/VLSM/Lib/NeList.v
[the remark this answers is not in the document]
traiansf: I removed them both; I was playing to see how/whether it works.

theories/VLSM/Lib/NeList.v
```diff
@@ -208,7 +195,8 @@ Lemma ne_list_concat_min_length
 Proof.
   induction l; cbn; [by lia |].
   rewrite app_length.
-  (* for some strange reason lia doesn't work directly, so... this hack: *)
+  (* by specialize (ne_list_min_length a); unfold ne_list_length; lia. *)
+  (* for some strange reason the line above does not work, so... this hack: *)
   assert (Hle : forall a b c, a >= 1 -> b >= c -> a + b >= S c) by lia.
   by apply Hle; [apply ne_list_min_length |].
```
wkolowski, line 201: I still don't like it. You should prove a lemma like `ne_list_length l = length (ne_list_to_list l)` and use it here.

theories/VLSM/Lib/NeList.v
```diff
@@ -217,42 +205,41 @@ Qed.
...
-| nel_cons h (Some t) => tl' h t (NonEmptyList_ind' A P hd' tl' t)
+| NEL_cons h None => hd' h
+| NEL_cons h (Some t) => tl' h t (NonEmptyList_ind' A P hd' tl' t)
 end.
 
 Fixpoint NonEmptyList_to_ne_list `(l : NonEmptyList A) : ne_list A :=
-  match l with
-  | nel_cons a None => ne_one a
-  | nel_cons a (Some l') => ne_cons a (NonEmptyList_to_ne_list l')
+  match (nel_tl l) with
```
wkolowski, line 234: Parentheses around `nel_tl l` not needed.

theories/VLSM/Lib/Preamble.v
```diff
@@ -194,63 +194,76 @@ Proof.
...
-    by exfalso; eapply Hnp; [done | lia |].
+    destruct (decide (find_least_among_helper  n = m')); [by lia |].
+    by exfalso; eapply find_least_among_helper_minimal with n m'; [| lia |].
 Qed.
 
+Lemma find_least_among_is_minimal
+  : minimal_among le P find_least_among.
+Proof.
+  destruct (find_least_among_helper_is_minimal bound) as [[[_ ?] ?] Hmin'].
+  cbn in *.
```
wkolowski, line 257: This does nothing.

theories/VLSM/Core/VLSMProjections/VLSMStutteringEmbedding.v
```diff
@@ -0,0 +1,725 @@
...
+    in Hpxt as (is_s & tr_s & Htr & Heqs).
+  replace (state_project s) with
+    (finite_trace_last (state_project is_s)
+      (pre_VLSM_stuttering_embedding_finite_trace_project transition_item_project tr_s)).
+  - eapply valid_trace_forget_last, finite_valid_trace_from_to_app_split.
+    rewrite <- pre_VLSM_stuttering_embedding_finite_trace_project_app.
+    by apply basic_VLSM_stuttering_embedding_finite_valid_trace_init_to.
+  - subst s; apply pre_VLSM_stuttering_embedding_finite_trace_last;
+      [by apply basic_VLSM_stuttering_embedding_type |].
+    by eapply valid_trace_forget_last, finite_valid_trace_from_to_app_split, Htr.
```
wkolowski:
```suggestion
  - subst s; apply pre_VLSM_stuttering_embedding_finite_trace_last.
    + by apply basic_VLSM_stuttering_embedding_type.
    + by eapply valid_trace_forget_last, finite_valid_trace_from_to_app_split, Htr.
```

theories/VLSM/Core/VLSMProjections/VLSMStutteringEmbedding.v
```diff
@@ -0,0 +1,725 @@
...
+Qed.
+
+(* end hide *)
```
wkolowski, line 704: There's `end hide`, but no `begin hide`.

wkolowski (approved):
Looks good. I committed some suggestions.

Please address the remaining suggestions before merging.

theories/VLSM/Core/VLSMProjections/VLSMStutteringEmbedding.v
```diff
@@ -0,0 +1,725 @@
...
+
+Definition VLSM_stuttering_embedding_infinite_finite_trace_project
+  (Hsimul : VLSM_stuttering_embedding X Y state_project transition_item_project)
+  (s : Streams.Stream (vtransition_item X))
+  (Hfin : FinitelyManyBound (fun item => transition_item_project item <> []) s)
+  : list (vtransition_item Y)
+  := pre_VLSM_stuttering_embedding_infinite_finite_trace_project transition_item_project s Hfin.
+
+Context
+  (Hsimul : VLSM_stuttering_embedding X Y state_project transition_item_project).
```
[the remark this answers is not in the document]
traiansf: Well, I need the preceding 4 lemmas to depend on a `VLSM_stuttering_embedding`, so they carry it as an additional argument, hence it seemed not so nice to add it to the context.  Maybe I should add a section there (and a comment) to make things more clear.

PR #139, opened 2022-11-14
remove spec tactic with multiple arguments

palmskog:
As discussed with  @bmmoore and @wkolowski and @traiansf, I here remove the multiple-argument versions of the `spec` tactic. The single-argument version is kept.


wkolowski (approved):
Looks good.

bmmoore (approved):
Looks good. All the changes are very simple.

PR #141, opened 2022-11-21
Code review for #138

wkolowski:
Part 1 of code review for #138.

### DIFF
--- a/theories/VLSM/Lib/NeList.v
+++ b/theories/VLSM/Lib/NeList.v
@@ -6,7 +6,7 @@ Inductive ne_list (A : Type) : Type :=
 | ne_one : A -> ne_list A
 | ne_cons : A -> ne_list A -> ne_list A.
 
-Arguments ne_one {_} _ : assert.
+Arguments ne_one  {_} _ : assert.
 Arguments ne_cons {_} _ _ : assert.
 
 Infix ":::" := ne_cons (at level 60, right associativity).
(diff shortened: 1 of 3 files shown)


theories/VLSM/Lib/StreamExtras.v
```diff
@@ -794,13 +793,15 @@ Lemma stream_concat_prefix {A} (s : Stream (ne_list A)) n len
...
 Proof.
-  intros ->; revert s prefix; induction n; [done |]; intros []; cbn.
+  intros ->; revert s prefix.
+  induction n; [done |].
+  intros [n0 s]; cbn.
   rewrite app_length.
-  assert (length (ne_list_to_list n0) +
-  length (mjoin (List.map ne_list_to_list (stream_prefix s n)))
-  ≥ length (ne_list_to_list n0)) by lia.
+  assert (H :
```
traiansf, line 800: if you insist on giving it a name, could it be a "non-anonymous" one?

theories/VLSM/Lib/NeList.v
```diff
@@ -28,16 +28,14 @@ Definition ne_list_foldr {A B} (f : A -> B -> B) (b : B) (l : ne_list A) :=
     match l with
     | ne_one a => f a b
     | ne_cons a nel => f a (fold nel)
-    end
-  ) l.
+    end) l.
```
traiansf, line 31: I guess here (and below) we should probably apply Karl's previous suggestion and bind the fix into a `let`

traiansf (approved):

theories/VLSM/Lib/StreamExtras.v
wkolowski: Yes, feel free to commit some changes here. I chose `H` because that was the default name.

PR #143, opened 2022-11-24
Extracting some non-triviality-related results about reachable threshold from older CBC formalization.

traiansf:


theories/VLSM/Core/ReachableThreshold.v
```diff
@@ -0,0 +1,177 @@
...
+  that the threshold is reachable ([reachable_threshold]) when there exists a
+  set of validators whose combined weight passes the threshold.
+
+  In this module we prove that there exist a [potentially_pivotal] validator,
+  i.e., a validator which added to a set of validators tips over the threshold.
+*)
+
+Class ReachableThreshold V Cv `{Hm : Measurable V} `{FinSet V Cv} : Set :=
+{
+  threshold : {r | (r >= 0)%R};
```
wkolowski, line 16: Is this way of stating it necessary? I think it confuses `lra` in many places. A better way would be to do `threshold : R` and then have the condition `r >= 0` as a separate field.

traiansf: I don't think that there was any particular reason for doing it that way (unless to make it clear it's a positive real), but i think refactoring that deserves a separate PR as it will touch a lot of files

palmskog: I think the proper way to write this class in the Stdpp style would be something like:
```coq
Class ReachableThreshold V Cv `{Hm : Measurable V} `{FinSet V Cv} (threshold:R) : Prop :=
{
  reachable_threshold_geq_0 : (r >= 0)%R;
  reachable_threshold : exists (vs : Cv), (sum_weights vs > threshold)%R;
}
```
But if we want this, good to do in a separate PR.

palmskog (approved):
Looks good to me, but consider a followup PR with some changes to the `ReachableThreshold` class.

theories/VLSM/Core/ReachableThreshold.v
```diff
@@ -0,0 +1,177 @@
...
+  reachable_threshold : exists (vs : Cv), (sum_weights vs > proj1_sig threshold)%R;
```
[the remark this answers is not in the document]
traiansf: Although I agree with you about the threshold, I think I agree more with Karl about the `reachable_threshold` constraint.
The idea here is that `threshold` is indeed relevant for the class, while the set of validators whose combined weight exceeds the threshold is not relevant; its only purpose that it exists, and therefore we can show that a pivotal validator also exists.

PR #144, opened 2022-11-25
Refactor some stream lemmas

wkolowski:
This proof reasoned by cases separately on `n <= m` and `n = m`. I changed it to use `Nat.lt_total`.


palmskog (approved):
Fine by me.

PR #145, opened 2022-11-25
Remove lemma `nodup_append`

wkolowski:
I removed the lemma `nodup_append`, because `NoDup_app` from the standard library (or is it stdpp?) can be used instead. In a later PR, uses of this lemma in casper-cbc-proofs will also have to be fixed.

I also removed `set_diff_nodup'` because it is exactly the same as `set_diff_nodup`. @palmskog You should take a look to see whether the TODO in the comment above `set_diff_nodup'` is still relevant.

I also fixed indentation in one place, which is not related to this PR, but I didn't want to forget about it.

### DIFF
--- a/theories/VLSM/Lib/ListSetExtras.v
+++ b/theories/VLSM/Lib/ListSetExtras.v
@@ -448,11 +448,7 @@ Qed.
 Lemma set_diff_nodup' `{EqDecision A} (l l' : list A)
   : NoDup l -> NoDup (set_diff l l').
 Proof.
-  induction 1 as [|x l H H' IH]; simpl.
-  - by constructor.
-  - case_decide.
-    + by apply IH.
-    + by apply set_add_nodup, IH.
+  by apply set_diff_nodup.
 Qed.
 
 Lemma diff_app_nodup `{EqDecision A} : forall (s1 s2 : list A),
--- a/theories/VLSM/Lib/ListSetExtras.v
+++ b/theories/VLSM/Lib/ListSetExtras.v
@@ -461,9 +461,9 @@ Lemma diff_app_nodup `{EqDecision A} : forall (s1 s2 : list A),
   NoDup (set_diff s1 s2 ++ s2).
 Proof.
   intros.
-  apply nodup_append; [| done |].
+  apply NoDup_app; split_and!; [| | done].
   - by apply set_diff_nodup.
-  - by intros; apply (set_diff_elim2 a s1).
+  - by intros a; apply (set_diff_elim2 a s1).
 Qed.
 
 Lemma add_remove_inverse `{EqDecision X}:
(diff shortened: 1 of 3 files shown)


bmmoore (approved):
Looks good, if you resolve the TODO comment.

theories/VLSM/Lib/ListSetExtras.v
```diff
@@ -440,31 +440,15 @@ Proof.
...
-Lemma set_diff_nodup' `{EqDecision A} (l l' : list A)
-  : NoDup l -> NoDup (set_diff l l').
-Proof.
-  induction 1 as [|x l H H' IH]; simpl.
-  - by constructor.
-  - case_decide.
-    + by apply IH.
-    + by apply set_add_nodup, IH.
-Qed.
-
```
palmskog, line 457: @wkolowski you can see [here](https://github.com/coq/coq/blob/0f1327e33a63f6d6acb375036a3bc83c0f2262f5/theories/Lists/ListSet.v#L457-L462) that the version in Coq's stdlib is still buggy - it has the unnecessary `NoDup l'` premise. We would like to keep this correct version in our library until the issue is fixed in Stdlib. So please roll back this change.

palmskog (changes requested):
This is is fine except for removal of `set_diff_nodup'`. Please roll that back and I'll approve.

theories/VLSM/Lib/ListSetExtras.v
wkolowski: @palmskog I got confused because there already is a lemma `set_diff_nodup` with the type we want, but it comes from `VLSM.Lib.StdppListSet`, not from the standard library. And it can be used to prove the lemma I removed simply as `by apply set_diff_nodup`, so it seems `set_diff_nodup'` is not needed after all.

I am still confused... please take a look at the most recent commit.

palmskog: @wkolowski here's how it was supposed to work: 
- the `StdppListSet.v` implements exactly the interface of Coq's Stdlib ListSet interface. This includes using precisely the lemma statements that are in Coq's stdlib
- the `set_diff_nodup'` was supposed to be *our* local implementation of `set_diff_nodup` with improvements applied, with an eye towards changing the version in Coq's stdlib.

However, in a PR with no stated relation to list sets, `set_diff_nodup` in `StdppListSet.v` was apparently changed in 6ec5e49267be2. I think we should roll back this change so that `StdppListSet.v`  faithfully reflects the Stdlib ListSet. However, there's no need to do this in this PR.

palmskog (approved):
This is fine by me now.

theories/VLSM/Lib/ListSetExtras.v
wkolowski: I'm still a bit confused, as the changes you linked are just to the proof of `set_diff_nodup`, but its statement had only one `NoDup` premise at that point already.

PR #148, opened 2022-12-05
More stdpp-like ReachableThreshold

traiansf:


theories/VLSM/Core/ReachableThreshold.v
```diff
@@ -4,26 +4,26 @@ From VLSM.Lib Require Import RealsExtras Measurable ListExtras StdppListSet.
...
-  that the threshold is reachable ([reachable_threshold]) when there exists a
+  that the threshold is reachable ([rt_reachable]) when there exists a
   set of validators whose combined weight passes the threshold.
 
   In this module we prove that there exist a [potentially_pivotal] validator,
   i.e., a validator which added to a set of validators tips over the threshold.
 *)
 
-Class ReachableThreshold V Cv `{Hm : Measurable V} `{FinSet V Cv} : Set :=
+Class ReachableThreshold V Cv (threshold : R) `{Hm : Measurable V} `{FinSet V Cv} : Prop :=
```
palmskog, line 14: This looks good, but I see that the class `Measurable` looks like this:
```coq
Class Measurable (V : Type) : Type :=
{
  weight : V -> pos_R;
}.
```
In a future PR, I think we should make this class as well be more Stdpp-like:
```coq
Class Measurable (V : Type) : Type := weight : V -> pos_R.
```
The difference is subtle but is actually important for proof automation (there is no constructor for an instance with this approach).

theories/VLSM/Core/Equivocation/LimitedMessageEquivocation.v
```diff
@@ -312,16 +316,16 @@ Qed.
...
   (Hfull : forall i, message_dependencies_full_node_condition_prop (IM i) message_dependencies)
   (no_initial_messages_in_IM : no_initial_messages_in_IM_prop IM)
   (can_emit_signed : channel_authentication_prop IM id sender)
-  (Htracewise_basic_equivocation : BasicEquivocation (composite_state IM) index Ci
-    := equivocation_dec_tracewise IM (fun i => i) sender)
-    (tracewise_not_heavy := @not_heavy _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ Htracewise_basic_equivocation)
-  : forall is s tr, strong_trace_witnessing_equivocation_prop IM id sender is tr (Cm := Ci) ->
+  (Htracewise_basic_equivocation : BasicEquivocation (composite_state IM) index Ci threshold
+    := equivocation_dec_tracewise IM threshold (fun i => i) sender)
+  (tracewise_not_heavy := not_heavy (1 := Htracewise_basic_equivocation) (Cv := Ci))
```
palmskog, line 327: A bit unconventional to use both argument number and argument name here. Isn't there a name instead of the `1`? If not, then I guess it's fine by me.

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/LimitedStateEquivocation.v
```diff
@@ -134,11 +135,11 @@ Proof.
...
     unfold not_heavy, Equivocation.not_heavy,
       equivocation_fault, Equivocation.equivocation_fault; simpl.
-    destruct threshold as [t Ht]; simpl; apply Rge_le.
-    pose proof (Heqv_is := equivocating_indices_equivocating_validators IM is).
+    etransitivity; [| by apply Rge_le, rt_positive].
+    pose proof (Heqv_is := equivocating_indices_equivocating_validators IM threshold is).
     rewrite equivocating_indices_initially_empty in Heqv_is by done.
     simpl in Heqv_is; apply sum_weights_empty in Heqv_is.
-    by rewrite <- Heqv_is in Ht.
+    by cbv in *; lra.
```
palmskog, line 142: It's always shaky performance-wise to do `tac in *`, due to the unknown size of the context (which may change over time). Maybe there is a more focused `cbv`?

palmskog (approved):
This looks good to me, only have some minor comments.

theories/VLSM/Core/Equivocation/LimitedMessageEquivocation.v
traiansf: No, it's not, because that argument is actually the class instance and `not_heavy` is a (defined) field;

theories/VLSM/Core/ReachableThreshold.v
traiansf: Would it make sense for Measurable to declare `weight : V -> R` and have an additional constraint saying `weight` is positive for every input?

traiansf: I've made the change you suggested and it did not break anything, so I've committed it in this PR.

palmskog: I think the `weight : V -> pos_R` is just fine. The approach from Stdpp would have required a separate `Prop` typeclass with the property that `weight` always gives positive output.

wkolowski: @traiansf @palmskog My guess is that having `weight : V -> R` with a constraint saying the result is positive would be an improvement with respect to automation, since `lra` seems not to understand `pos_R`. I recall writing `by destruct (Measurable.weight x); simpl in *; lra` in a few places in the PR about using `lra` more.

theories/VLSM/Core/ELMO/ELMO.v
```diff
@@ -1436,7 +1437,7 @@ Lemma ELMO_initial_state_not_heavy :
 Proof.
   intros s Hs.
   unfold ELMO_not_heavy, not_heavy.
-  replace (equivocation_fault s) with 0%R; [by destruct threshold; apply Rge_le |].
+  replace (equivocation_fault s) with 0%R; [by apply Rge_le, rt_positive |].
```
wkolowski, line 1440: This goal requires proving `(0 <= threshold)%R`, but `rt_positive : (threshold >= 0)%R`. Maybe it would be better if `rt_positive` was of type `(0 <= threshold)%R`?

theories/VLSM/Core/Equivocation/LimitedMessageEquivocation.v
```diff
@@ -67,7 +68,7 @@ Proof.
     as [[[is His] ->] | (l & [s' om'] & om & [(_ & _ & _ & Hv) Ht])]; simpl.
   - exists ∅.
     + by intros v Hv; contradict Hv; apply Hno_initial_equivocation.
-    + by destruct threshold; simpl; apply Rge_le; rewrite sum_weights_empty.
+    + by rewrite sum_weights_empty; [apply Rge_le, rt_positive |].
```
wkolowski, line 71: Again, we need `apply Rge_le, rt_positive` because `rt_positive` uses `Rge` instead of `Rle`.

theories/VLSM/Core/Equivocation/LimitedMessageEquivocation.v
```diff
@@ -145,14 +147,14 @@ Definition tracewise_limited_equivocation_vlsm_composition : VLSM message :=
 
 Lemma full_node_limited_equivocation_valid_state_weight s
   : valid_state_prop tracewise_limited_equivocation_vlsm_composition s ->
-    LimitedEquivocationProp IM is_equivocating s.
+    LimitedEquivocationProp (Cv := Ci) IM threshold is_equivocating s.
 Proof.
-  apply limited_equivocation_valid_state.
+  eapply limited_equivocation_valid_state; [done |].
```
wkolowski, line 152: This is some kind of regression, since `eapply` was not needed before. Can you investigate?

theories/VLSM/Core/Equivocation/MsgDepLimitedEquivocation.v
```diff
@@ -87,11 +88,11 @@ Qed.
...
 Lemma coeqv_limited_equivocation_state_not_heavy s
   : valid_state_prop coeqv_limited_equivocation_vlsm s ->
-    (sum_weights (state_annotation s) <= proj1_sig threshold)%R.
+    (sum_weights (state_annotation s) <= threshold)%R.
 Proof.
   induction 1 using valid_state_prop_ind.
-  - destruct s, Hs as [_ ->], threshold; cbn in *.
-    by rewrite sum_weights_empty; [apply Rge_le |].
+  - destruct s, Hs as [_ ->]; cbn in *.
+    by rewrite sum_weights_empty; [apply Rge_le, rt_positive |].
```
wkolowski, line 95: Again, `apply Rge_le, rt_positive`.

theories/VLSM/Core/Equivocation/MsgDepLimitedEquivocation.v
```diff
@@ -473,7 +478,7 @@ Lemma msg_dep_fixed_limited_equivocation_witnessed
         tr).
 Proof.
   repeat split; [..| by apply Htr].
-  - by apply coeqv_limited_equivocation_state_not_heavy,
+  - by eapply coeqv_limited_equivocation_state_not_heavy,
```
wkolowski, line 481: Again, `eapply` is needed instead of `apply`.

theories/VLSM/Core/Equivocation/WitnessedEquivocation.v
```diff
@@ -147,7 +149,7 @@ Lemma input_valid_transition_reflects_trace_witnessing_equivocation_prop
 Proof.
   apply finite_valid_trace_init_to_last in Htr as Hlst.
   intros v; split; simpl in *; rewrite Hlst in *.
-  - by intros Hv; eapply equivocating_validators_is_equivocating_tracewise_iff with (Cm := Cm).
```
wkolowski, line 150: Have you checked all places where `with (Cm := Cm)` occurs whether it can be removed now?

theories/VLSM/Core/Equivocators/Composition/LimitedEquivocation/LimitedStateEquivocation.v
```diff
@@ -134,11 +135,11 @@ Proof.
   - subst s.
     unfold not_heavy, Equivocation.not_heavy,
       equivocation_fault, Equivocation.equivocation_fault; simpl.
-    destruct threshold as [t Ht]; simpl; apply Rge_le.
-    pose proof (Heqv_is := equivocating_indices_equivocating_validators IM is).
+    etransitivity; [| by apply Rge_le, rt_positive].
```
wkolowski, line 138: And again `apply Rge_le, rt_positive`.

theories/VLSM/Core/ReachableThreshold.v
```diff
@@ -4,26 +4,26 @@ From VLSM.Lib Require Import RealsExtras Measurable ListExtras StdppListSet.
...
 {
-  threshold : {r | (r >= 0)%R};
-  reachable_threshold : exists (vs : Cv), (sum_weights vs > proj1_sig threshold)%R;
+  rt_positive : (threshold >= 0)%R;
```
wkolowski, line 16: My guess is that this would help.

```suggestion
  rt_positive : (0 <= threshold)%R;
```

theories/VLSM/Core/ReachableThreshold.v
```diff
@@ -114,18 +113,18 @@ Qed.
...
   exists v,
     v ∈ vs /\
-    (sum_weights (vs ∖ {[v]}) <= proj1_sig threshold)%R.
+    (sum_weights (vs ∖ {[v]}) <= threshold)%R.
 Proof.
   intros vss Hvss.
   destruct (pivotal_validator_extension ∅ vss)
     as (vs & Hincl & Hvs & v & Hv & Hvs').
-  - by rewrite sum_weights_empty; [destruct threshold; apply Rge_le |].
+  - by rewrite sum_weights_empty; [destruct Hrt; lra |].
```
wkolowski, line 127: This time the pattern was hidden, but there it is:

```suggestion
  - by rewrite sum_weights_empty; [apply Rge_le, rt_positive |].
```

wkolowski (approved):
Looks ok, but please take a look at the suggestions. I did not commit any of them.

theories/VLSM/Core/ReachableThreshold.v
palmskog: I don't think out-of-the-box compatibility with `lra` should dictate these design choices. Better if we look for pre-processing tools or just regular tactics down the line that can turn a `pos_R` into an `R` and a constraint.

theories/VLSM/Core/Equivocation/LimitedMessageEquivocation.v
traiansf: I think it's because before threshold was a member of the class so the class  was automatically inferred

theories/VLSM/Core/Equivocation/WitnessedEquivocation.v
traiansf: I've checked some (most?). it's still a bit mystifying for me when it manages to automatically infer and when not.

PR #149, opened 2022-12-07
Stepwise oracles

bmmoore:
Adjust the definitions of HasBeenSentCapability and HasBeenReceivedCapability so the class directly uses the stepwise oracle correctness properties. The trace-based correctness properties for `has_been_sent` and `has_been_received` are  still available but now defined as lemmas rather then directly being class members.


theories/VLSM/Core/Equivocation.v
```diff
@@ -362,6 +369,151 @@ Definition no_traces_have_message_prop
...
+  (oracle : state_message_oracle)
+  : Prop :=
+{
+  oracle_no_inits :
+    forall (s : vstate vlsm),
+      initial_state_prop (VLSMMachine := vmachine vlsm) s ->
+      forall m, ~ oracle s m;
+  oracle_step_update :
+    forall l s im s' om,
+      input_valid_transition (pre_loaded_with_all_messages_vlsm vlsm) l (s, im) (s', om) ->
```
traiansf, line 417: I wonder whether it would make sense to use here `input_valid_transition_item`

theories/VLSM/Core/Equivocation.v
```diff
@@ -362,6 +369,151 @@ Definition no_traces_have_message_prop
...
+  constructor; [done |].
+  by intros; rewrite Hupdate, Heqv.
+Qed.
+
+Lemma oracle_trace_props_from_stepwise
+  `(Horacle: oracle_stepwise_props selector oracle) :
+  oracle_tracewise_props selector oracle.
+Proof.
+  constructor; intros s Hs m.
+  - with_strategy opaque [iff] hnf.
```
traiansf, line 488: I haven't seen this tactic before in our codebase. Is there a specific reason to start using it?

theories/VLSM/Core/Equivocation.v
```diff
@@ -931,7 +1058,7 @@ Defined.
 
 End sec_stepwise_from_trace.
 
-(** ** Stepwise view of [HasBeenSentCapability]
+(** ** Stepwise view of [HasBeenSentCapability] TODO - move up with HasBeenSent
```
traiansf, line 1061: Is it OK to have a TODO in a coqdoc header, even title?

theories/VLSM/Core/Equivocation.v
```diff
@@ -945,44 +1072,15 @@ End sec_stepwise_from_trace.
...
-Defined.
-
 Lemma preloaded_has_been_sent_stepwise_props
       [message : Type]
       (vlsm: VLSM message)
       `{HasBeenSentCapability message vlsm}
       (seed : message -> Prop)
       (X := pre_loaded_vlsm vlsm seed):
-  has_been_sent_stepwise_props (vlsm := X) (has_been_sent vlsm).
+  has_been_sent_stepwise_prop (vlsm:=X) (has_been_sent vlsm).
```
traiansf, line 1081: spaces around `:=`

theories/VLSM/Core/Equivocation.v
```diff
@@ -1860,11 +1860,11 @@ Lemma preloaded_composite_has_been_received_stepwise_props
   (constraint : composite_label IM -> composite_state IM * option message -> Prop)
   (seed : message -> Prop)
   (X := pre_loaded_vlsm (composite_vlsm IM constraint) seed)
-  : has_been_received_stepwise_props (vlsm := X) composite_has_been_received.
+  : has_been_received_stepwise_prop (vlsm:=X) composite_has_been_received.
```
traiansf, line 1863: spaces around `:=`

traiansf (approved):
looks reasonable to me.

theories/VLSM/Core/ELMO/UMO.v
```diff
@@ -108,7 +108,7 @@ Qed.
 Instance HasBeenSentCapability_UMOComponent
   (i : Address) : HasBeenSentCapability (UMOComponent i).
 Proof.
-  eapply HasBeenSentCapability_from_stepwise with (fun s m => m ∈ sentMessages s)
+  apply Build_HasBeenSentCapability with (fun s m => m ∈ sentMessages s)
```
palmskog, line 111: It's a bit unfortunate that we have to explicitly reference the constructor names here and just below - this is typically not considered good practice. However, there might be no obvious better thing to do here.

theories/VLSM/Core/Equivocation.v
```diff
@@ -333,15 +336,19 @@ Proof.
...
 (**
-  Checks if all [valid_trace]s leading to a certain state contain a certain message.
-  The [message_selector] argument specifies whether we're looking for received or sent
-  messages.
+  The oracle should check if all valid traces leading to the state contains certain message.
+  The [message_selector] argument checks whether a single transition contains the
+  message, and can be used to check for received messages or sent messages.
 
-  Notably, the [valid_trace]s over which we are iterating belong to the preloaded
+  Notably, the traces we are considering are any valid in the preloaded
```
palmskog, line 343: A bit more clear:
```suggestion
  Notably, the traces we are considering are any traces valid in the preloaded
```

theories/VLSM/Core/Equivocation.v
```diff
@@ -362,6 +369,151 @@ Definition no_traces_have_message_prop
...
+          forall m, selector m item <-> selector' m item)
+      : oracle_stepwise_props selector' oracle.
+Proof.
+  destruct Horacle as [Hinits Hupdate].
+  constructor; [done |].
+  by intros; rewrite Hupdate, Heqv.
+Qed.
+
+Lemma oracle_trace_props_from_stepwise
+  `(Horacle: oracle_stepwise_props selector oracle) :
```
palmskog, line 484: I guess @wkolowski can see to this, but this way of handling variables with the backtick is inconsistent with what we have just above.

theories/VLSM/Core/Equivocation.v
palmskog: @bmmoore please make the TODO a separate, standalone non-coqdoc comment, like this:
```coq
(* TODO: move up with HasBeenSent *)
```

palmskog (approved):
Looks reasonable to me, but please consider the comments before merging.

theories/VLSM/Core/Equivocation.v
```diff
@@ -362,6 +369,151 @@ Definition no_traces_have_message_prop
...
+    split.
+    + intros Hholds s0 tr Htr.
+      by eapply (oracle_initial_trace_update Horacle).
+    + apply pre_loaded_with_all_messages_valid_state_prop,
+        valid_state_has_trace in Hs as (start & tr & Htr).
+      intro H; specialize (H start tr Htr).
+      by eapply oracle_initial_trace_update.
+  - with_strategy opaque [iff] hnf.
+    unfold not_oracle.
+    hnf_in selected_message_exists_in_no_preloaded_trace; simpl.
```
wkolowski, line 499:
```suggestion
  - red; unfold not_oracle, selected_message_exists_in_no_preloaded_trace,
      specialized_selected_message_exists_in_no_trace.
```

theories/VLSM/Core/Equivocation.v
```diff
@@ -388,25 +540,61 @@ Definition has_not_been_received_prop : state_message_oracle -> state -> message
...
+  := oracle_step_update _ _ has_been_sent_stepwise_props.
+
+Definition has_been_sent_tracewise_prop
+      (has_been_sent_pred : state_message_oracle) : Prop :=
+  oracle_tracewise_props (field_selector output) has_been_sent_pred.
+
+Lemma has_been_sent_tracewise_props `{HasBeenSentCapability} :
+  has_been_sent_tracewise_prop has_been_sent.
+Proof.
+  by exact (oracle_trace_props_from_stepwise has_been_sent_stepwise_props).
```
wkolowski, line 578: I think we should use `apply` instead of `exact`.

```suggestion
  by apply oracle_trace_props_from_stepwise, has_been_sent_stepwise_props.
```

theories/VLSM/Core/Equivocation.v
```diff
@@ -434,6 +622,7 @@ Lemma has_been_sent_consistency
   : selected_messages_consistency_prop (field_selector output) s m.
 Proof.
   split; [| by apply consistency_from_valid_state_proj2].
+  pose proof has_been_sent_tracewise_props.
```
wkolowski, line 625: What's the point of adding this line?

```suggestion
```

theories/VLSM/Core/Equivocation.v
```diff
@@ -640,118 +866,31 @@ Qed.
 
 End sec_simple.
 
-#[global] Hint Mode HasBeenSentCapability - ! : typeclass_instances.
-#[global] Hint Mode HasBeenReceivedCapability - ! : typeclass_instances.
+Arguments oracle_stepwise_props {message} {vlsm} message_selector oracle.
+Arguments oracle_no_inits {message} {vlsm} {message_selector} {oracle}.
+Arguments oracle_step_update {message} {vlsm} {message_selector} {oracle}.
```
wkolowski, line 871: I like aligning `Arguments`, as it makes it easier to compare across definitions.

```suggestion
Arguments oracle_stepwise_props {message} {vlsm} message_selector oracle.
Arguments oracle_no_inits       {message} {vlsm} {message_selector} {oracle}.
Arguments oracle_step_update    {message} {vlsm} {message_selector} {oracle}.
```

theories/VLSM/Core/Equivocation.v
```diff
@@ -1242,30 +1287,16 @@ Lemma has_been_directly_observed_sent_received_iff
...
-      by exists is, tr, Htr.
-  - intros [Hreceived | Hsent]; apply Hall; intros is tr Htr.
-    + apply proper_received in Hreceived; [| done].
-      apply Exists_or. left.
-      by eapply Hreceived.
-    + apply proper_sent in Hsent; [| done].
-      apply Exists_or. right.
-      by eapply Hsent.
+  revert m.
+  induction Hs using valid_state_prop_ind; intro m.
```
wkolowski, line 1291: Apparently, the `revert` is not needed.

theories/VLSM/Core/Equivocation.v
```diff
@@ -1242,30 +1287,16 @@ Lemma has_been_directly_observed_sent_received_iff
...
+  - split.
+    + by intros ?%has_been_directly_observed_no_inits.
+    + by intros [?%has_been_received_no_inits | ?%has_been_sent_no_inits].
```
wkolowski, line 1294: My guess is that @palmskog won't like this one.

theories/VLSM/Core/Equivocation.v
```diff
@@ -1441,7 +1441,7 @@ Definition ComputableSentMessages_has_been_sent
   `{EqDecision message}
   : HasBeenSentCapability vlsm.
 Proof.
-  eapply HasBeenSentCapability_from_stepwise; cycle 1.
+  econstructor; cycle 1.
   - by apply csm_computable_oracle.
   - by typeclasses eauto.
```
wkolowski, line 1446: @bmmoore @palmskog @traiansf I wonder: maybe we should prefer chaining over `cycle`, like in the suggestion?

```suggestion
  econstructor; [| by apply csm_computable_oracle].
  by typeclasses eauto.
```

theories/VLSM/Core/Equivocation.v
```diff
@@ -1778,12 +1778,11 @@ Qed.
...
   (X := composite_vlsm IM constraint)
-  : has_been_sent_stepwise_props (vlsm := X) composite_has_been_sent.
+  : has_been_sent_stepwise_prop (vlsm := X) composite_has_been_sent.
 Proof.
   unfold has_been_sent_stepwise_props.
   pose proof (composite_stepwise_props
-                (fun i => has_been_sent_stepwise_from_trace
-                            (IM i)))
+                (fun i => has_been_sent_stepwise_props (IM i)))
        as [Hinits Hstep].
```
wkolowski, line 1786: @bmmoore @palmskog @traiansf This is a very interesting case, because `pose proof ... as [Hinits Hstep]` works just fine, but `destruct ... as [Hinits Hstep]` doesn't. Any body knows why?

wkolowski (approved):
Looks ok. I opened a PR with suggestions (it also contains the various minor suggestions by Traian and Karl): #150

theories/VLSM/Core/Equivocation.v
```diff
@@ -362,6 +369,151 @@ Definition no_traces_have_message_prop
...
+*)
+
+Record oracle_stepwise_props
+  (message_selector : message -> transition_item -> Prop)
+  (oracle : state_message_oracle)
+  : Prop :=
+{
+  oracle_no_inits :
+    forall (s : vstate vlsm),
+      initial_state_prop (VLSMMachine := vmachine vlsm) s ->
```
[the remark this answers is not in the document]
bmmoore: I was keeping this from old code, but I think it probably doesn't actually help. The idea is to help unification. We have a reasonable amount of code that writes `initial_state_prop` instead of using `vinitial_state_prop` (especially in the early sections defining or just after defining a new VLSM), and that ends up inferred as something like `initial_state_prop (VLSM_machine:=composite_vlsm_machine ...)`. The idea was it might be easier to unify that against `initial_state_prop (VLSMMachine := ?vmachine) ?s` than against `vinitial_state_prop ?vlsm ?s`. But I think the argument actually being written `vmachine vlsm` means this unification problem is also not solved so automatically. (I think turning on an eta rule with "Primitive Projections" would let `vmachine ?vlsm` unify with a concrete machine, but eta would also let `vinitial_state_prop ?vlsm` unify against `initial_state_prop (VLSMMachine:=?machine)`).

theories/VLSM/Core/Equivocation.v
bmmoore: Perhaps. I thought we were using more components of the transition item, but it's actually just s', so maybe a `destination item` in the conclusion wouldn't be too heavy. 

The other issue is that we often have `input_valid_transition` hypotheses, so we would want to have a version of the property with this type anyway. So I think we should leave this one alone, and maybe add the `input_valid_transition_item` form as a lemma if it looks useful.

theories/VLSM/Core/Equivocation.v
bmmoore: I resorted to `with_strategy` because `hnf` doesn't take the usual "delta flags" form to tell an evaluation rule not to unfold something, like in `cbv -[iff]`.

theories/VLSM/Core/ELMO/UMO.v
bmmoore: It's unfortunate we couldn't use record construction syntax. I tried `{| |}` in many of these places, and had much typechecking trouble. In particular it seems it tries to typecheck the fields without using any context from looking at the expected type of the record, and `apply` on the constructor does. And `constructor` doesn't seem to have any way to pass your implementation of `has_been_sent`. I think `unshelve econstructor;[by exact (fun s m => m ∈ sentMessages s)|..].` could work, but by that point you might as well just write the constructor name.

theories/VLSM/Core/Equivocation.v
bmmoore: Because `pose` allows the result to have existential variables, and the constraint isn't fixed until we use the new hypotheses. `edestruct ... as [Hints Hstep]` works.

bmmoore: I don't know why it doesn't require `epose`

theories/VLSM/Core/Equivocation.v
```diff
@@ -208,6 +208,9 @@ Qed.
 Definition state_message_oracle
   := vstate vlsm -> message -> Prop.
 
+Definition not_oracle (o : state_message_oracle) : state_message_oracle :=
```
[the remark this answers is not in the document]
bmmoore: I went with `negate_oracle`

PR #151, opened 2022-12-09
Replace some Unicode characters with their ASCII counterparts.

wkolowski:
This is the equivalent of https://github.com/runtimeverification/casper-cbc-proofs/pull/542 for vlsm.


bmmoore (approved):
As with runtimeverification/casper-cbc-proofs#542 I'm not necessarily opposed to unicode in general, but these replacements all look good.

PR #154, opened 2022-12-12
switch to Docker-Coq-Action for CI

palmskog:
We switch to using CI based on [Docker-Coq-Action](https://github.com/coq-community/docker-coq-action).


F-WRunTime (approved):

palmskog:
@traiansf maybe we actually want to have CI for *any* branch, regardless if it's related to a PR, and regardless of whether a PR is targeting `master`? What do you think?

Jenkinsfile
```diff
@@ -1,64 +0,0 @@
...
-        //stage('Deploy Docs') {
-        //  steps {
-        //    sshagent(['2b3d8d6b-0855-4b59-864a-6b3ddf9c9d1a']) {
-        //      sh '''
-        //        echo "Trigger Doc Update"
-        //      '''
-        //    }
-        //  }
-        //}
-        stage('Update Dependents') {
```
F-WRunTime, line 52: I think we need to keep this section for now to update the dependent repositories

F-WRunTime (changes requested):

F-WRunTime:
> 

Yeah that is reasonable to do on public runners. I've used workflow-dpsatch for this in the past to enable manually running on any branch as well as setting up the trigger to watch on push to any branch.

palmskog:
@F-WRunTime When I re-introduced a slimmer `Jenkinsfile`, I tried removing the Dockerfile, but I get this error:
```
WorkflowScript: 1: Missing required section "agent" @ line 1, column 1.

   pipeline {

   ^
```
What agent should I use if I want to avoid using any Dockerfile?

Jenkinsfile
```diff
@@ -1,38 +1,13 @@
 pipeline {
-  agent {
```
F-WRunTime, line 2: Needs this section and the Dockerfile returned to run properly

palmskog: But can't we use some standard Docker container from the Docker registry for this? The current Dockerfile does a ton of stuff that would take time to clean out.

F-WRunTime: We can probably simplify it. Let me take a look for what's needed

palmskog: I got something now that at least doesn't error out in the Jenkins workflow. Maybe you can re-review and check if we can do something even better.

.github/workflows/docker-action.yml
```diff
@@ -0,0 +1,27 @@
+name: Docker CI
+
+on:
+  push:
```
F-WRunTime, line 4: on: 
   push: 

^^ Above will capture pushing to any branch you can also add if you want to run manually when no changes are needed but want to see something run. Would look like the following

```
on:
   push:
   workflow_dispatch: 
```

F-WRunTime:
For the update-deps we can give this a try in a new file > release.yml
```
name: "Master Push"
on:
  push:
    branches:
      - master

jobs:
  release:
    name: 'Publish Release'
    runs-on: ubuntu-latest
    environment: production
    steps:
      - name: 'Check out code'
        uses: actions/checkout@v3

      - name: 'Update dependents'
        env:
          JENKINS_DEVOPS_TOKEN: ${{ secrets.JENKINS_DEVOPS_TOKEN }}
        run: |
          curl --fail 'https://ci.runtimeverification.com/jenkins/buildByToken/buildWithParameters' \
            --data job=Devops/master \
            --data token=$JENKINS_DEVOPS_TOKEN \
            --data UPDATE_DEPS=true \
            --data UPDATE_DEPS_REPO=runtimeverification/vlsm \
            --data UPDATE_DEPS_VERSION="$(git rev-parse HEAD)"
```

palmskog:
@F-WRunTime you mean, we try this `release.yml` as a replacement for both `Jenkinsfile` and `Dockerfile`?

F-WRunTime:
> 

Yes this would trigger on a merge to master to run the udpate-deps from the devops repository and upstream repositories should get the updated master branch as a PR.

.github/workflows/docker-action.yml
```diff
@@ -0,0 +1,27 @@
+name: Docker CI
```
F-WRunTime, line 1: For consistency with our repositories can we update this to 'Test PR' and the name of the yaml file to test-pr.yml

F-WRunTime:
I will update the checks to no longer wait for Jenkins

F-WRunTime (approved):

PR #156, opened 2022-12-13
Speculative VLSMs

wkolowski:


theories/VLSM/Core/Speculation.v
```diff
@@ -0,0 +1,97 @@
...
+
+Context
+  {message : Type}
+  (X : VLSM message).
+
+Inductive SpeculativeLabel : Type :=
+| Simulate  : SpeculativeLabel
+| Rollback  : SpeculativeLabel
+| Commit    : SpeculativeLabel
+| Speculate : list (vlabel X * option message) -> SpeculativeLabel.
```
traiansf, line 24: It might make sense to use here a plan from https://github.com/runtimeverification/vlsm/blob/master/theories/VLSM/Core/Plans.v
This would help connect with existing results about plans (e.g., valid plan, plan execution)

theories/VLSM/Core/Speculation.v
```diff
@@ -0,0 +1,97 @@
...
+
+Inductive SpeculativeState : Type :=
+| Underlying  : vstate X -> SpeculativeState
+| Speculative :
+    vstate X -> vstate X -> list (vlabel X * option message) -> list (option message) ->
```
traiansf, line 29: wouldn't a record serve you better here, allowing to give proper names to the components of a speculative state?

traiansf (approved):
Looks reasonable to me. I don't like much the name `Underlying` (seems too pretentious), but I don't care that much about names ...

denisadiaconescu (approved):
Looks good!

palmskog (commented):
As a general rule, I think we want at least some simple sanity-checking lemmas when new definitions are introduced. This would be mainly to gain confidence that definitions don't have unexpected corner cases and are not empty.

Also, is there an obvious reason the `Inhabited` definition is not an `Instance`? I'd separate the content that lives in `Type` from the stuff in `Prop` and only define the latter with tactics.

theories/VLSM/Core/Speculation.v
wkolowski: I think a record doesn't make much sense here, as it's a two-constructor type. And names can be given even without a record, just using a different syntax (but then they'll only be used to generate default names for `destruct` and the like).

wkolowski:
@palmskog I made `Inhabited` into an `Instance`. Regarding the second point, I looked over the codebase and lots of instances of `Inhabited` are defined using tactics, so I would keep it as is. I agree that we should have some lemmas to "test" new definitions, but I feel like in this case the lemmas are not that obvious to produce (at least for me), so I guess I'll merge it and open a separate PR for the lemmas.

PR #157, opened 2022-12-16
construct Inhabited typeclass members without tactics

palmskog:
The `Inhabited` typeclass lives in the sort `Type`, not in `Prop`. Hence, its members have computational meaning and should normally not be constructed using tactics (which can easily make terms blow up). Here is the standard approach to separate `Type` terms from `Prop` terms.


wkolowski (commented):
I have proposed to apply this kind of refactoring consistently, it's in PR #158. It also fixes the build failure from this PR caused by the name change.

wkolowski (approved):
You should probably change the name of the PR to reflect the wider scope of the changes.

PR #159, opened 2022-12-16
Add `Proof` after each `Next Obligation`

wkolowski:


theories/VLSM/Lib/Preamble.v
```diff
@@ -667,5 +667,6 @@ Definition sum_project_right {A B : Type} (x : A + B) : option B :=
 Program Definition not_lt_plus_dec {m n} (Hnlt : ~n < m) : {k | k + m = n} :=
   exist _ (n - m) _.
 Next Obligation.
+Proof.
   by cbn; lia.
 Qed.
```
palmskog, line 672: Might as well compress.
```suggestion
Proof. by cbn; lia. Qed.
```

palmskog (approved):
Looks good to me.

theories/VLSM/Lib/Preamble.v
wkolowski: @palmskog I was wondering about this kind of compression. Maybe we should always have `Proof` and `Qed` on separate lines? This would make the structure more predictable and easier to parse for the kind of hacky scripts we might want to write...

PR #160, opened 2022-12-16
Speculative VLSM lemmas

wkolowski:


theories/VLSM/Core/Speculation.v
```diff
@@ -62,8 +62,8 @@ Definition speculative_valid
 match sl, ssim with
 | Speculate plan, (Actual _, None) =>
     plan <> []
-| Execute, (Speculative _ current (Build_plan_item lbl msg :: _) _, None) =>
-    vvalid X lbl (current, msg)
+| Execute, (Speculative _ current (Build_plan_item lbl msg :: _) _, msg') =>
+    vvalid X lbl (current, msg) /\ msg = msg'
```
palmskog, line 66: @wkolowski @traiansf this and what is just above looks like a specification change, maybe you can add some coqdoc comment to explain the rationale of these particular match branches?

theories/VLSM/Core/Speculation.v
```diff
@@ -92,4 +92,185 @@ Definition speculative_vlsm : VLSM message :=
...
+      (Actual s2, None)).
+  {
+    apply input_valid_transition_Commit_nil.
+    by eapply input_valid_transition_destination.
+  }
+  constructor; [| done].
+  constructor; [| done].
+  constructor; [| done].
+  constructor; [| done].
+  constructor.
```
palmskog, line 245: This looks pretty repetitive, maybe just `repeat (constructor; try done)` would make more sense?

theories/VLSM/Core/Speculation.v
```diff
@@ -92,4 +92,185 @@ Definition speculative_vlsm : VLSM message :=
...
+    let exec   := @Build_transition_item _ speculative_vlsm_type
+                    Execute iom (Speculative start current [] [oom]) None in
+    let output := @Build_transition_item _ speculative_vlsm_type
+                    Commit None (Speculative start current [] []) oom in
+    let commit := @Build_transition_item _ speculative_vlsm_type
+                    Commit None (Actual current) None in
+      [spec; exec; output; commit]
+end.
+
+Fixpoint speculative_lift_trace
```
palmskog, line 181: Could we get some short documentation for these definitions? Not at all obvious to me what they are doing. This one looks like it might be better written by a fold, e.g., `fold_right`.

palmskog (approved):
Looks reasonable, but please consider the comments.

theories/VLSM/Core/Speculation.v
wkolowski: Fixed.

theories/VLSM/Core/Speculation.v
wkolowski: I don't think `speculative_lift_trace` fits into `fold_right`, as it's not only consuming the trace, but also the `start` state has to be updated in each step.

theories/VLSM/Core/Speculation.v
traiansf: The coqdoc would sound something like: asking that the `Execute` transition receives the message being executed from the speculative plan ensures that the message is _valid_ (in the sense of being produc-eable through a valid trace) for the `speculative_vlsm`; in particular, if the `speculative_vlsm` is further constrained, messages which are valid for the underlying vlsm might not be valid anymore for the constrained `speculative_vlsm`

theories/VLSM/Core/Speculation.v
traiansf: Although I agree with @palmskog that everything can be written as a `fold_right`, I don't clearly see what the benefit of using `fold_right` would be in this particular case (one would still need to write an append lemma which wouldn't be much simpler to prove by using the fold_right_app result).

That is, unless we decide to make all list recursions folds from now on :-)

theories/VLSM/Core/Speculation.v
```diff
@@ -92,4 +92,193 @@ Definition speculative_vlsm : VLSM message :=
...
+  (start : vstate X) (tr : list (vtransition_item X))
+  : list (vtransition_item speculative_vlsm) :=
+match tr with
+| [] => []
+| ti :: tr' =>
+    speculative_lift_transition_item start ti ++ speculative_lift_trace (destination ti) tr'
+end.
```
bmmoore, line 131: This could use `flat_map` or an equivalent

theories/VLSM/Core/Speculation.v
```diff
@@ -92,4 +92,193 @@ Definition speculative_vlsm : VLSM message :=
...
+
+Lemma speculative_lift_trace_app :
```
bmmoore, line 133: This would just be `flat_map_app` if `speculate_lift_trace` was defined with `flat_map`.

theories/VLSM/Core/Speculation.v
```diff
@@ -92,4 +92,193 @@ Definition speculative_vlsm : VLSM message :=
...
+      (Actual s2, None)).
+  {
+    apply input_valid_transition_Commit_nil.
+    by eapply input_valid_transition_destination.
+  }
+  repeat (constructor; try done).
+  by eapply input_valid_transition_destination.
+Qed.
+
+Lemma finite_valid_trace_init_to_speculative_lift_trace :
```
bmmoore, line 257: This seems like it should be easier to derive from `speculative_lift_input_valid_transition`, but maybe we don't have appropriate lemmas about concatenating a sequence of `finite_valid_trace_from_to`.

bmmoore (approved):
Looks ok.

The definition of `speculative_lift_trace` could be simpler using library functions.

theories/VLSM/Core/Speculation.v
wkolowski: I don't see a way of using `flat_map` here without doing some higher-order juggling.

theories/VLSM/Core/Speculation.v
wkolowski: I thought it was as simple as that, but it seems what was needed was `finite_valid_trace_init_to_rev_strong_ind`, the strongest induction principle for `finite_valid_trace_init_to` that we have. I will try to look for a better proof in a later PR with more lemmas.

theories/VLSM/Core/Speculation.v
bmmoore: Right, the start and destination don't fit into flat_map. It's closer to something about the transitive closure of relations, but  `transition_item` doesn't actually contain `start` so it those don't fit either.

PR #161, opened 2022-12-19
port to Coq 8.16, add CI job for 8.16

palmskog:
Minimal changes to support both Coq 8.15 and 8.16. We start to test continuously with both versions, then we can drop 8.15 after a while.

### DIFF
--- a/theories/VLSM/Core/ELMO/ELMO.v
+++ b/theories/VLSM/Core/ELMO/ELMO.v
@@ -1437,8 +1437,9 @@ Lemma ELMO_initial_state_not_heavy :
 Proof.
   intros s Hs.
   unfold ELMO_not_heavy, not_heavy.
-  replace (equivocation_fault s) with 0%R; [by apply rt_positive |].
-  by symmetry; apply sum_weights_empty, ELMO_initial_state_equivocating_validators.
+  replace (equivocation_fault s) with 0%R.
+  - by apply (rt_positive (H6 := H6)).
+  - by symmetry; apply sum_weights_empty, ELMO_initial_state_equivocating_validators.
 Qed.
 
 Lemma ELMO_not_heavy_send_message :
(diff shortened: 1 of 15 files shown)


theories/VLSM/Core/ELMO/ELMO.v
```diff
@@ -1437,8 +1437,9 @@ Lemma ELMO_initial_state_not_heavy :
 Proof.
   intros s Hs.
   unfold ELMO_not_heavy, not_heavy.
-  replace (equivocation_fault s) with 0%R; [by apply rt_positive |].
-  by symmetry; apply sum_weights_empty, ELMO_initial_state_equivocating_validators.
+  replace (equivocation_fault s) with 0%R.
+  - by apply (rt_positive (H6 := H6)).
```
wkolowski, line 1441: I know you don't like `eapply`, but maybe it'd be better here?

theories/VLSM/Core/ReachableThreshold.v
```diff
@@ -98,10 +98,13 @@ Proof.
...
-        apply elements_proper.
-        intro x.
-        by rewrite elem_of_difference, !elem_of_list_to_set, set_remove_iff, elem_of_singleton.
+        etransitivity.
+        -- symmetry.
+           by apply (elements_list_to_set (H6 := H6)), set_remove_nodup.
+        -- apply elements_proper.
+           intro x.
+           rewrite elem_of_difference, !elem_of_list_to_set.
+           by rewrite set_remove_iff, elem_of_singleton.
```
wkolowski, line 107: Isn't there a way to save the level of nesting here?

wkolowski (approved):

theories/VLSM/Core/ELMO/ELMO.v
palmskog: How would `eapply` help here? Just dropping the `(H6 := H6)` is not going to work. The only difference from before is that an `apply` becomes longer and I therefore split into several bullets.

theories/VLSM/Core/ReachableThreshold.v
palmskog: I don't think the nesting level is a problem in this small proof. Some more bullets can also be helpful to catch failing proofs.

PR #165, opened 2023-01-18
update documentation for Coq 8.16 and new Platform release

palmskog:
Since the [new Coq 8.16 Platform](https://github.com/coq/platform/releases/tag/2022.09.1) is out, I update the documentation on installing VLSM. In particular, itauto is now part of the Platform, so those who use Platform installers and scripts can build VLSM right away.

8.15 also has known soundness bugs, which, although they are very unlikely to affect us, points to recommending 8.16 going forward.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -14,7 +14,6 @@ jobs:
       matrix:
         image:
           - 'coqorg/coq:8.16-ocaml-4.14-flambda'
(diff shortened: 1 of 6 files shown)


wkolowski (approved):

traiansf (approved):

palmskog:
I decided that we should fully switch to 8.16 and Stdpp 1.8.0, since there will otherwise have to be developer effort to support both 8.15 and 8.16. However, then we will need some assistance to get this merged.

PR #169, opened 2023-02-06
Remove redundant imports

TeodorescuIoan:
These minimizations were obtained by calling the appropriate absolutize imports and minimize import scripts and noticing which imports were removed.

### DIFF
--- a/theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces/FixedSetByzantineTraces.v
@@ -1,7 +1,7 @@
 From Cdcl Require Import Itauto. #[local] Tactic Notation "itauto" := itauto auto.
 From stdpp Require Import prelude finite.
 From Coq Require Import FunctionalExtensionality.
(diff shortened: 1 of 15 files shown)


theories/VLSM/Core/ByzantineTraces.v
```diff
@@ -1,5 +1,4 @@
 From stdpp Require Import prelude.
-From VLSM.Lib Require Import Preamble.
```
palmskog, line 2: I think we made a decision somewhere to keep the `Preamble` import here since we want everyone to have the same environment. So maybe you can roll back this particular line, @TeodorescuIoan?

palmskog (approved):
This looks good, I just have a one-line revision request. After this is addressed you can merge.

(But don't forget to provide also some documentation of the toolchain in a different PR)

traiansf (approved):
Looks good to me, too.
I wonder whether it might be a good idea to also add the helping scripts you wrote to help you with the task; or they weren't much of a help in the end?
Of course, if you do that you would need to properly document them, so only do it if you think it would help when we would want to re-run this process

PR #170, opened 2023-02-07
Improve building instructions

TeodorescuIoan:
Added install instructions for Fedora

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -25,36 +25,46 @@ sudo apt-get update
 
 ### Install opam (unless already installed)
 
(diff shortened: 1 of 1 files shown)


wkolowski (commented):
I think it's not a good idea to intersperse instructions for different OSes. It would be better if each kind of installation (Ubuntu/Fedora/Nix/whatever) was kept in a separate section.

palmskog:
Since the OS/distribution specific parts are quite few, i think I found a compromise.

palmskog (approved):
Let's get this merged now as a small incremental improvement.

PR #179, opened 2023-03-07
Improve axiom diagnostics script

wkolowski:
Some axioms were incorrectly filtered out from the output by the script, but only their types were supposed to be filtered out. This made e.g. the uses of classical logic not be detected.

I also filtered out `PrimInt63` and `Uint63` related axioms from the output so that they don't clutter it.

### DIFF
--- a/scripts/axiom-diagnostics.sh
+++ b/scripts/axiom-diagnostics.sh
@@ -74,15 +74,20 @@ coqc $COQLIBS "$tmp/tmp.v" \
 | \
 sed -r \
 `# Remove axiom types written inline.` \
--e '/ : .*/d' \
+-e 's/(.*) : .*/\1/' \
 `# Remove axiom types written multiline.` \
 -e '/^  .*/d' \
 `# Remove compilation details.` \
 | \
- grep -vE '(^COQ)|(^make\[))' \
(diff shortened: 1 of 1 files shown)


scripts/axiom-diagnostics.sh
```diff
@@ -73,16 +73,21 @@ coqc $COQLIBS "$tmp/tmp.v" \
...
--e '/ : .*/d' \
 `# Remove axiom types written multiline.` \
 -e '/^  .*/d' \
+`# Remove axiom types written inline.` \
+-e 's/([^:]*) : .*/\1/' \
 `# Remove compilation details.` \
 | \
- grep -vE '(^COQ)|(^make\[))' \
-`# Remove redundant lines and indent axioms listings ` \
+grep -vE '(^COQ)|(^make\[))' \
```
TeodorescuIoan, line 82: You can actually remove this as it does not filter the details anyways

wkolowski: Removed.

wkolowski:
@TeodorescuIoan Ok, I removed that line. Can you approve?

TeodorescuIoan (approved):

PR #184, opened 2023-03-16
Filter out comments for axiom diagnostics

TeodorescuIoan:


palmskog (approved):
Looks reasonable, but please add some comment about that this (seemingly) only works for non-nested comments. That is, Coq allows comments like:
```coq
(* comment (* goes here *) and here *)
```

wkolowski (approved):
I tested it and it seems to work. I second Karl's request, you should add a comment to explain what the added part is doing.

PR #185, opened 2023-03-17
Remove some uses of `Existing Instance`

wkolowski:
I always found the command `Existing Instance` suspicious - I think you shouldn't need to use it. It turns out we can get rid of quite a few of them while keeping everything working just like before.

### DIFF
--- a/theories/VLSM/Core/ELMO/ELMO.v
+++ b/theories/VLSM/Core/ELMO/ELMO.v
@@ -204,8 +204,6 @@ Proof.
     by intros [(-> & -> & m2 & Hrecv & Hadr & Hincomp) |]; econstructor.
 Qed.
 
-#[local] Existing Instance list_exist_dec.
-
 #[export]
 Instance local_equivocators_full_obs_dec : RelDecision local_equivocators_full_obs.
 Proof.
@@ -213,7 +211,11 @@ Proof.
   induction ol using addObservation'_rec.
   - by right; inversion 1.
   - apply (Decision_iff (iff_Symmetric _ _ (lefo_alt _ _ _))).
-    by typeclasses eauto.
+    apply Decision_or; [| done].
+    apply Decision_and; [done |].
+    apply Decision_and; [by typeclasses eauto |].
+    apply list_exist_dec.
+    by intro; typeclasses eauto.
 Defined.
 
 Definition local_equivocators_full (s : State) : Address -> Prop :=
--- a/theories/VLSM/Core/Equivocation/TraceWiseEquivocation.v
+++ b/theories/VLSM/Core/Equivocation/TraceWiseEquivocation.v
@@ -307,7 +307,7 @@ Context
   `{RelDecision _ _ is_equivocating_tracewise_no_has_been_sent}
   .
 
-#[local] Program Instance equivocation_dec_tracewise
+#[export] Program Instance equivocation_dec_tracewise
   : BasicEquivocation (composite_state IM) validator Cv threshold :=
   {
     state_validators := fun _ => list_to_set (enum validator);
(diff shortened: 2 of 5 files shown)


theories/VLSM/Core/ELMO/ELMO.v
```diff
@@ -204,16 +204,18 @@ Proof.
...
-#[local] Existing Instance list_exist_dec.
-
 #[export]
 Instance local_equivocators_full_obs_dec : RelDecision local_equivocators_full_obs.
 Proof.
   intros ol a.
   induction ol using addObservation'_rec.
   - by right; inversion 1.
   - apply (Decision_iff (iff_Symmetric _ _ (lefo_alt _ _ _))).
-    by typeclasses eauto.
```
traiansf, line 216: I don't see why it's better to replace an `Existing Instance` with a more elaborate proof script.

theories/VLSM/Core/Equivocation/TraceWiseEquivocation.v
```diff
@@ -307,7 +307,7 @@ Context
   `{RelDecision _ _ is_equivocating_tracewise_no_has_been_sent}
   .
 
-#[local] Program Instance equivocation_dec_tracewise
+#[export] Program Instance equivocation_dec_tracewise
```
traiansf, line 310: It's not clear one wants to export that; we have multiple definitions of Basic Equivocation for a composition and we might want to be able to select which one we want to use in a given context; sometimes maybe both would be active.

traiansf (commented):
Was this a priority? I guess we should take Karl's advice, to compile a list of issues related to refactorings (or to curate the one we already have) and then to periodically meet and set priorities.

wkolowski:
@traiansf I don't think it was a priority. "Meeting and setting priorities" sounds like a very professional, well-thought-out strategy and whatnot, but I think it's often much faster and more efficient to open this kind of quick PR and gather feedback during code review.

I did it mostly because I never used `Existing Instance` despite having projects with hundreds of instances, so I was very curious if this is something that is really necessary or just a sign of something fishy going on. You can treat this as my career development and learning at work :)

My conclusion is that often `Existing Instance` is a code smell (the most obvious cases being places which had `Lemma` instead of `Instance`), but there were a few places in which I was not able to get rid of it and I'm not completely sure why.

@palmskog I'm also looking forward to your opinion on this.

theories/VLSM/Core/ELMO/ELMO.v
```diff
@@ -204,16 +204,18 @@ Proof.
...
+    apply Decision_or; [| done].
+    apply Decision_and; [done |].
+    apply Decision_and; [by typeclasses eauto |].
+    apply list_exist_dec.
+    by intro; typeclasses eauto.
```
bmmoore, line 218: It would be easiser to just give `typeclasses eauto` the help locally:
```suggestion
    pose proof @list_exist_dec.
    by typeclasses eauto.
```

bmmoore (approved):
Note that `typeclasses eauto` will also use existing definitions if you add them to the hypotheses with `pose proof` instead of declaring an `Existing Instance`.
I put a suggestion for your one proof change.
(If we had many lemmas that we wanted to optionally use, we could make a HintDb for them and use the `typeclasses eauto with typeclass_instances our_hints ...` form).

I also checked that `casper-cbc-proofs` still builds with this, so removing the non-`#[local]` instances is fine.

theories/VLSM/Core/ELMO/ELMO.v
wkolowski: Thanks, I didn't know you can do that! I tried `typeclasses eauto with list_exist_dec`, but it didn't work.

bmmoore: Yeah, `typeclasses eauto` does have `with $HINTDBS` option but it doesn't support the `using $LEMMAS` clause that `auto` and `eauto` have.

PR #188, opened 2023-03-17
Put all annotations inline instead of above

wkolowski:
@traiansf @palmskog @bmmoore I propose to add the convention that locality annotations are placed on the same line as the command they annotate. This makes life easier when using `grep` or `sed`.


traiansf (commented):
I have no opinion on this; if you think it's needed for scripts and Karl agrees, then go on and merge once he approves.

CODING_CONVENTIONS.md
```diff
@@ -158,6 +158,21 @@ Record simp_lv_event : Type := SimpObs
...
+### Locality annotations
+
+Locality annotations should be put on the same line as the command they annotate. This makes it easier to process them with line-based tools, like `grep` or `sed`.
+
+Good:
+```coq
+#[export] Instance compare_eq_dec {A} `{CompareStrictOrder A} : EqDecision A.
+```
+
+Bad:
```
palmskog, line 170: I don't like the terminology "bad", and there are many places where annotations will make the line extremely long.
```suggestion
Not recommended:
```

CODING_CONVENTIONS.md
```diff
@@ -158,6 +158,21 @@ Record simp_lv_event : Type := SimpObs
...
+```coq
+#[export]
+Instance compare_eq_dec {A} `{CompareStrictOrder A} : EqDecision A.
+```
+
```
palmskog, line 175: Can you please come up with some rule of thumb for how to handle long lines with annotations and `Instance`, for example? It's natural to place the linebreak after the annotation.

palmskog (commented):
Changes to code look reasonable, but please address comments.

CODING_CONVENTIONS.md
bmmoore: That seems to be forced if we want line-based tools to be able to connect the annotation with the definition name.

bmmoore (approved):
approving

CODING_CONVENTIONS.md
```diff
@@ -158,6 +158,35 @@ Record simp_lv_event : Type := SimpObs
...
+#[export] Instance compare_eq_dec {A} `{CompareStrictOrder A} : EqDecision A.
+```
+
+Not recommended:
+```coq
+#[export]
+Instance compare_eq_dec {A} `{CompareStrictOrder A} : EqDecision A.
+```
+
+In case you need to split the line because it's too long, keep the annotation on the same line as the command and the name. When splitting the above example:
```
palmskog, line 176:
```suggestion
In case you need to split a long line, keep the annotation on the same line as the command and the name. When splitting the above example:
```

palmskog (approved):
Fine by me now, but see one suggestion.

CODING_CONVENTIONS.md
wkolowski: Done.

PR #190, opened 2023-03-20
Keep all definitions in axiom analysis output

TeodorescuIoan:


wkolowski (approved):
It seems to work fine, although I only tested it on a single directory.

wkolowski:
@TeodorescuIoan Also, don't forget to also update the script in casper-cbc-proofs: commit it to the update dependency PR.

PR #191, opened 2023-03-20
Add group by option to make axioms

TeodorescuIoan:
The options can be activated by passing GROUPBYMOD=true to the make axioms command, for example:
```bash
make GROUPBYMOD=true axioms
```

Obs. if group by module option is active the writing will no longer happen in real time. The contents will be dumpted at the end.

### DIFF
--- a/scripts/axiom-diagnostics.sh
+++ b/scripts/axiom-diagnostics.sh
@@ -1,35 +1,35 @@
 #!/bin/sh
 
-GROUPBYFLAG=0
 
-if test -n "$1"
+if test -n "$4"
 then
-   GROUPBYFLAG=$1
+   GROUPBYFLAG=$4
 fi
 
-# If argument $2 is not present, print usage info and exit.
-if test -z "$2"
+# If argument $1 is not present, print usage info and exit.
+if test -z "$1"
 then
   echo "Usage:"
   echo "make axioms"
   echo "make axioms path=path_to_source_directory"
   echo "make axioms path=path_to_source_directory keep_tmp=true"
+  echo "make axioms path=path_to_source_directory keep_tmp=true GROUPBYMOD=true"
   exit
 fi
 
-# Nice name for argument $2.
-COQLIBS=$2
+# Nice name for argument $1.
+COQLIBS=$1
 
-# If argument $3 is not present, set it to current directory.
-if test -z $3
+# If argument $2 is not present, set it to current directory.
+if test -z $2
 then
   dir=.
 else
-  dir=$3
+  dir=$2
 fi
 
-# If argument $4 is set to true, then keep it. Otherwise set it to false.
-if [[ $4 ]] && [ $4 == true ]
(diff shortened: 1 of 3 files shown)


wkolowski (changes requested):
I think you shouldn't change the existing argument order - the new one should come at the end.

Also, can `make` parse arguments by name, instead of positionally? If so, the script should be updated to indicate that.

I also see that you have a huge duplicated chunk of code in both of the `if` branches - this should be abstracted into a function or at least factored out to the front of the `if` somehow.

palmskog (commented):
I agree with @wkolowski to put the new option last. Also, recall that in Makefiles you can do stuff like `MY_VAR =? "default"` to always give a value to something, regardless of whether the user provides the option. I think this should be used here.

palmskog:
@wkolowski can you please re-review this? The argument order and other things now look good to me.

scripts/axiom-diagnostics.sh
```diff
@@ -1,12 +1,19 @@
 #!/bin/sh
 
+
+if test -n "$4"
+then
+   GROUPBYFLAG=$4
```
wkolowski, line 6: Please use only 2 spaces on indentation to stay consistent with the rest of the file. Also, why is the variable name different here than in the makefile?

```suggestion
  group_by_mod=$4
```

scripts/axiom-diagnostics.sh
```diff
@@ -90,7 +97,10 @@ awk '{if (lastLine=="") {lastLine=$0;next} if ($0 ~ /Closed under the global con
 | \
 sed -r \
 -e '/PrimInt63/d' \
--e '/Uint63/d'
+-e '/Uint63/d' \
+`# If the user passed GROUPBYMOD=true then group the listings by module name.` \
+| \
+if [ "$GROUPBYFLAG" = "true" ]; then sed -E -e 's/\t/|/' -e '/^\|/!s/()\..*$//' | awk 'BEGIN {lastModule=""} {if ($0 ~ /^\|/) { data[lastModule][$0] = 1 } else { lastModule=$0 }} END {for (key in data) { print key; for (axiom in data[key]){ print "\t"axiom; } print "\n"; }}' | sed -E 's/\|//'; else tail -n +1; fi
```
wkolowski, line 103:
```suggestion
if [ "$group_by_mod" = "true" ]; then sed -E -e 's/\t/|/' -e '/^\|/!s/()\..*$//' | awk 'BEGIN {lastModule=""} {if ($0 ~ /^\|/) { data[lastModule][$0] = 1 } else { lastModule=$0 }} END {for (key in data) { print key; for (axiom in data[key]){ print "\t"axiom; } print "\n"; }}' | sed -E 's/\|//'; else tail -n +1; fi
```

wkolowski (approved):
I tested it and it works fine, but please address my suggestions before merging (and make sure it still works after you do all the renames).

PR #192, opened 2023-03-22
Remove uses of the Opaque command

wkolowski:


traiansf:
I think the purpose was to prevent `simpl` from simplifying them.

traiansf:
Since I don't have a good feeling about it, I'm adding Karl as a reviewer.

wkolowski:
@palmskog My point is that there's no reason (as hinted by the first silly use of `Opaque` right before closing a section). The manual says `Opaque` persists only till the end of a module, so the reason for using it must be some lemma about `finite_valid_trace` from `VLSM.v`. I stepped through all of them and there is nothing extraordinary going on. It never gets unfolded, and it gets hit by `simpl` maybe 2 or 3 times, but that of course doesn't unfold it either.

palmskog:
@wkolowski I won't approve the PR unless it uses the MathComp style locking.

bmmoore:
@wkolowski For any PR removing something that possibly has a good purpose besides merely letting proofs pass, it would be good to include a reference to the PR adding the code, and if possible evidence (beyond the build passing) that the removal doesn't hurt the possible objective (also in code added after the initial PR).

In this case a benchmark would at least show that there is no performance hit to a `compute` or `cbv` or such from removing the `Opaque`. And as you point out, both are limited to their section so you just need to build the one file. Some of the timing targets in https://coq.inria.fr/refman/practical-tools/utilities.html#timing-targets-and-performance-testing could help, and testing with a `sleep` (https://github.com/tchajed/coq-tricks/blob/master/src/Sleep.v) could help establish that the times are stable enough to at least see a 1sec difference.

Because of further investigation I would rather remove this completely, but I do agree with @palmskog that it's inappropriate to *only* remove the `Opaque` and leave other vestiges of the (failed attempt at) locking.

It turns out both `Opaque` were added in my own runtimeverification/casper-cbc-proofs#193.

I don't remember what the idea was with `finite_valid_trace`.

With `finite_trace_last` it was intended as a refactoring aid, and I probably forgot to remove it or perhaps intentionally left it in in imitation of MathComp without any concrete need. I was extracting `finite_trace_last` from bare uses of the expression `List.last (List.map destination tr) s`, and to also remove raw reasoning about `last` and `map` from proofs about traces by adding lemmas. I knew `apply`ing a map lemma could still succeed by unfolding the definition (and wasn't so clear on `rewrite`) and added Opaque to deliberately try to break them.

You point the `Opaque` actually does nothing, and I notice all the changed places were `rewrite`, so we should probably check that nothing breaks with a properly `#[global]` opaque, and *then* remove it.

bmmoore:
Oh, actually I think `Opaque`-ing `finite_valid_trace` may have been intended as a similar refactoring aid. But we've completely abandoned the idea of treating the `_init` versions of trace properties as opaque predicates in favor of `destruct`ing them with intro patterns whenever we want.

bmmoore (approved):
Ok, thanks @wkolowski. Those are no reason not to remove the `Opaque` and finish this PR, just a sign that my original refactoring that extracted the `finite_trace_last` definition was incomplete.

I didn't intent to make it impossible to unfold `finite_trace_last` and reason on the definition, just to provide enough lemmas directly stated in terms of `finite_trace_last` that it wouldn't often be necessary to do that.

wkolowski:
@palmskog So, since I and Brandon agree this should be removed, are you willing to approve now?

palmskog (approved):
OK, since @bmmoore was the author of this and has approved, I'm willing to approve. I still think the general solution is to use proper locking, possibly via Elpi.

PR #193, opened 2023-03-29
Refactor `VLSM_eq`

wkolowski:
The goal of this PR is to change the definition of `VLSM_eq X Y` to `VLSM_incl X Y /\ VLSM_incl Y X` and remove all redundancies that result from this.


bmmoore (approved):
This definition does seem to be equivalent.
Much better this way.

PR #194, opened 2023-03-30
Refactor ListExtras

wkolowski:
This PR is possibly best reviewed commit-by-commit.

Explanations for each commit:
- I removed the function `is_member` which was sloppily written decision procedure for list membership because we already have something else for this purpose (`inb` I think)
- I removed `null_dec` which was an instance of `Decision` for `l = []` because we can just use decidable equality for that
- I removed the function `compareb` because it was an unused function which overlaps with the instance of `EqDecision` for types that are `StrictlyComparable`
- ~~I rewrote the lemma `Forall_filter` to be `Lemma` proved by `induction` instead of `Fixpoint`~~
- I refactored the function `complete_prefix` which was needlessly complicated
- I removed the lemma `in_fast` which was unused and is easily proved in-place anyway
- I removed the lemma `in_one_element_decompositions_iff` which was a wrapper around similar lemma using `elem_of`
- I removed `ForAllSuffix1` and `ExistsSuffix1`  because they are equivalent to `Forall` and `Exists` and not used anywhere
- I removed lemmas `empty_nil` and `elem_of_empty_nil` because stdpp already has such a lemma


traiansf (commented):
Hm... Interesting. I always thought that you need `EqDecision `a`  to get `EqDecision (list a)`; I vaguely remember `decide (l = [])` didn't used to work some time ago. But then, again, I might be remember it wrong.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -595,14 +595,14 @@ Proof.
     by specialize (IHl n0 eq_refl n3 eq_refl); lia.
 Qed.
 
-Fixpoint Forall_filter
+Lemma Forall_filter
```
traiansf, line 589: if it ends with Defined, shouldn't it be a `Definition`?

wkolowski:
@traiansf Indeed you need `EqDecision A` for `EqDecision (list A)`, otherwise what would you do in the case `[a] = [b]`? Not sure what makes you think about this, as nothing in this PR touches on `EqDecision` for `list A`.

theories/VLSM/Lib/ListExtras.v
wkolowski: If it ends with `Defined`, maybe it shouldn't be defined in proof mode?

This lemma looks fishy, together with `Forall_hd` and `Forall_tl`, which are also definitions living in `Prop` that end in `Defined`, and the fishiness stems from the fact that the function `list_annotate` relies upon the computatioanl behavior of these things.

I can revert this change if you want, but this should be dealt with at some point.

bmmoore:
It doesn't change the decision to remove, but the point is that `Decision (l = [])` for `list A` does not need (to search for) an instance `EqDecision A`. 

Anyway, stdpp has `list_eq_nil_dec`, so it is now redundant. (haven't checked when it was introduced).

theories/VLSM/Lib/ListExtras.v
palmskog: This results lives in `Prop`, so it's fine by to change it from  `Fixpoint` to `Lemma`, but there is basically not much gain in changing anything with these results. If we can change it to use `Qed`, that's usually a good thing, but there may be some proof that requires `Defined` somewhere.

palmskog (commented):
This looks reasonable, but please take into account Brandon's comment. It's nearly always better to defer to Stdpp whenever we can.

theories/VLSM/Lib/ListExtras.v
wkolowski: @traiansf @palmskog Ok, I reverted this change and added a TODO that reminds us to deal with these `Defined` proofs.

wkolowski:
@bmmoore @palmskog I'm not sure if I understand correctly, but if stdpp has an instance for `Decision (l = [])` and I removed our custom one, then there is nothing more to be done, right?

theories/VLSM/Lib/ListExtras.v
```diff
@@ -12,9 +12,9 @@ From VLSM.Lib Require Import Preamble.
 Lemma has_last_or_null {S} (l : list S)
   : {l' : list S & {a : S | l = l' ++ (a :: nil)}} + {l = nil} .
 Proof.
-  destruct (decide (l = [])).
+  destruct l as [| h t].
```
bmmoore, line 15: The `destruct (decide (l = []))` should also still work here, and picks up the `list_eq_nil_dec` from stdpp. So if it stopped working for you something is strange. But changing to `destruct l` for simplicity is fine.

```suggestion
  destruct l.
```
would be enough - the names are not used.

bmmoore (approved):
Nothing else needs to be changed about the `Decision (l = [])` instances, but see the question in ListExtras.

theories/VLSM/Lib/ListExtras.v
wkolowski: It worked fine, but I decided that maybe it's better to just use `destruct` instead of going through instances and `decide` for such a simple thing.

PR #196, opened 2023-03-31
Move instances related to list comparison to the Preamble

wkolowski:


palmskog (commented):
This change is fine by me (to consolidate all `StrictlyComparable`-related definitions), but I'll leave to @traiansf to approve.

traiansf (approved):
I'm not sure Preamble is the perfect destination for it, but SortedLists isn't either, so...

PR #197, opened 2023-03-31
Remove `Listing`

wkolowski:
Some time ago we got rid of most of uses of `Listing` and replaced them with `Finite` from stdpp. Now time has come to get rid of the remaining uses.

Changes in this PR:
- I deleted the file `FinFunExtras.v` which was about `Listing` lemmas
- I deleted all other `Listing` lemmas
- I proved a lemma about `map_option` that will be needed to fix proofs in casper-cbc-proofs


theories/VLSM/Lib/ListExtras.v
```diff
@@ -1000,6 +1000,40 @@ Proof.
...
+    by firstorder congruence.
+Qed.
+
+Lemma NoDup_map_option :
+  forall {A B : Type} (f : A -> option B) (l : list A),
+    (forall a1 a2 : A, is_Some (f a1) -> is_Some (f a2) -> f a1 = f a2 -> a1 = a2) ->
+      NoDup l -> NoDup (map_option f l).
+Proof.
+  induction l as [| h t]; cbn; [by constructor |].
+  intros Hinj Hnd; inversion Hnd; subst; clear Hnd.
```
bmmoore, line 1029: You can avoid the `inversion` by doing the `induction` on directly on the `NoDup` fact  `Hnd` instead.

bmmoore (approved):
Good for `Listing`, but your proof of `elem_of_map_option` worsens another redundancy with `stdpp` - our `map_option` with `omap`/`list_omap`.

Please reduce the proof to `elem_of_list_omap` as suggested.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1000,6 +1000,40 @@ Proof.
...
+    + intros [-> | Hin]; [by exists h; itauto |].
+      apply IHt in Hin as (x & Hinx & Hfx).
+      by exists x; itauto.
+    + intros (x & [-> | Hin] & Hfx); rewrite IHt.
+      * by left; congruence.
+      * by right; exists x.
+  - setoid_rewrite elem_of_cons.
+    rewrite IHt.
+    by firstorder congruence.
+Qed.
```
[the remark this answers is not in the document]
wkolowski: Thanks a lot. I guessed this lemma should have been proven somewhere in stdpp, but I was not able to find it as we are using `map_option` instead of `omap`.

theories/VLSM/Lib/ListExtras.v
wkolowski: Fixed.

PR #199, opened 2023-04-04
Relate VLSM_incl with a VLSM_embedding by eq_rect

bmmoore:
In case we have a `(X Y : VLSM message)` and `vtype X = vtype Y`, then the `VLSM_type` components are propositionally equal but not definitionally equal.
We can relate a `VLSM_embedding` that projects with `eq_rect` to a `VLSM_incl` claim, by constructing a new VLSM that is propositionally equal to `Y` but has a `VLSM_type` definitionally equal to `vtype X`, by using the equality on `vtype`s to change the type in `Y`'s `VLSM_machine`.

The technique at the beginning of the proof for using the equality might be useful other places where we have an `eq` claim between parts of VLSMs but not (yet) definitional equality
```
  destruct X as [XT XM]; simpl in *; set (X := {| vmachine := XM |}) in *.
  destruct Y as [YT YM]; simpl in *; set (Y := {| vmachine := YM |}) in *.
  subst YT.
```

I didn't find a good example, but made a minor example of using the new lemma in `FixedSetByzantineTraces.v` (we could completely drop the old `VLSM_incl_part` stuff, and the `VLSM_incl` notation just a `VLSM_incl_alt` over `eq_refl`).

@wkolowski Please take a look and see if any ideas from this are useful


wkolowski:
From what I remember, this approach is not much of an improvement, so I'll close this PR. I opened a related issue so that the topic of improving `VLSM_incl` and related notions is not lost.

PR #200, opened 2023-04-04
Remove most uses of `In`, `incl` and `elem_of_list_In`. Cleanup `map_option` a bit.

wkolowski:
Changes:
- Replaced `In` with `∈` everywhere.
- Replaced `incl` with `⊆` everywhere.
- Removed most uses of `elem_of_list_In`.
- Removed quite a few lemmas about `In` and `incl` that were either not used or had equivalents in stdpp.
- Refactored some lemmas about `In` and `incl` to be about `∈` and `⊆`.
- Also refactored some lemmas about `map_option` and `list_sum` that were standing in my way.


wkolowski (commented):
Self-review.

theories/VLSM/Core/Equivocators/FullReplayTraces.v
```diff
@@ -168,7 +168,7 @@ Proof.
       | _ =>  full_replay_state i
       end)).
   {
-    intros Hcut; specialize (Hcut _ (incl_refl _) ltac:(apply NoDup_enum)).
+    intros Hcut; specialize (Hcut _ ltac:(done) ltac:(apply NoDup_enum)).
```
traiansf, line 171: I remember it was agreed that we will try to limit the usages of `ltac:` arguments.
I know that there is already one there; but I was wondering whether we could factor both of them out.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -145,56 +140,44 @@ Lemma in_not_in : forall A (x y : A) (l : list A),
   x <> y.
 Proof. by itauto congruence. Qed.
 
-Definition inb {A} (Aeq_dec : forall x y : A, {x = y} + {x <> y}) (x : A) (xs : list A) :=
-  if in_dec Aeq_dec x xs then true else false.
+Definition inb `{EqDecision A} (x : A) (xs : list A) :=
```
traiansf, line 143: do we still use `inb`?
And... couldn't this be refactored into using `decide (x \in xs)` 
Maybe for a subsequent refactoring task :-)

traiansf (approved):
Looks good to me.

theories/VLSM/Core/Equivocators/FullReplayTraces.v
wkolowski: Fixed.

PR #204, opened 2023-04-14
Remove module StdppListFinSet

wkolowski:
This is follow-up to #203. The module was almost entirely unused, so I removed it.


theories/VLSM/Core/SubProjectionTraces.v
```diff
@@ -1,7 +1,7 @@
 From VLSM.Lib Require Import Itauto.
 From Coq Require Import FunctionalExtensionality Lia.
 From stdpp Require Import prelude finite.
-From VLSM.Lib Require Import Preamble ListExtras StdppListSet StdppExtras StdppListFinSet.
+From VLSM.Lib Require Import Preamble ListExtras StdppListSet StdppExtras. (*  StdppListFinSet. *)
```
traiansf, line 4: You probably want to remove it, not just comment it out.

traiansf (approved):

theories/VLSM/Core/SubProjectionTraces.v
wkolowski: Fixed.

PR #205, opened 2023-04-14
Export `elem_of_dec_slow` from Preamble

wkolowski:
For some reason this instance was not exported by stdpp, which caused us to frequently write things like `pose @elem_of_dec_slow` or `apply elem_of_dec_slow`.


traiansf (approved):
My guess is that it wasn't exported especially because it's "slow"...
But since we don't care about program extraction, it makes sense for us to have it exported.

PR #209, opened 2023-04-17
Remove almost all lemmas from module RealsExtras

wkolowski:
Most lemmas in RealsExtras were unused.


bmmoore (approved):
Looks okay, if these are also unused in runtimeverification/casper-cbc-proofs.
These results all look like they could be easily proved by `lra` or similar tactics.

But do you know why we had these lemmas?
The only thing that comes to mind is that sometime I like to avoid writing out big terms by using something like `rewrite (some_lem x y)` even if the goal resulting from `replace (... x + ... y ...)%R with (... y * ... x ...)%R` can be solved by a single tactic.
But these results don't look like they would be especially useful for that either.

traiansf:
> But do you know why we had these lemmas?

I think we did use them with rewrite at some point. We did use to have similar results for natural numbers before we decided on using Lia.

I guess we were in the process of learning and felt like automation was hiding things away...

PR #210, opened 2023-04-17
Remove module RealsExtras

wkolowski:
Given that the only remaining thing in RealsExtras is a single instance of `Transitive` for `Rle`, it may make sense to get rid of the module entirely. The price for this is that we need to change uses of `transitivity` and `etransitivity` to the more ugly `eapply Rle_trans`.


bmmoore (commented):
How about moving the instance somewhere else instead of going without?

No modules would gain a transitive dependency on `Reals` if the instance moved into `Measurable`.

On the other hand, declaring the instance seems sort of like setting a global option to me, more than making a new declaration, and it might be a bit surprising to have that in `Measure.v`.

It would fit fine in `Preamble.v`, but we probably don't want everything to indirectly `Require Reals`?

wkolowski:
@bmmoore Ok, since it seems we don't want to remove the module entirely, I'm closing this PR for now.

PR #212, opened 2023-04-18
Remove module Temporal and move its contents to StreamExtras

wkolowski:
The module Temporal was quite misleading, because it sounds like it's about something to do with temporal logic, but actually it was all about streams and stream operations.

Summary of changes:
- I deleted the definition of `Eventually` and `Forever`, because they are identical to `Exists` and `ForAll` for streams
- I changed all the definition and lemmas talking about `Eventually` and `Forever` to use `Exists` and `ForAll`, including renaming them
- I refactored the lemmas a bit (e.g. I split the proof of `stabilization` into lemmas)
- I moved all this stuff to the module StreamExtras and deleted the module Temporal


bmmoore (approved):
Looks good.

We named it like this because these are the stream properties needed to
define the semantics of temporal logic operators, and we didn't have much else
on streams then.
Having a small section in `StreamExtras` for more specifically temporal stuff and otherwise unifying the names is all good.

PR #215, opened 2023-04-20
Change `VLSMType` from Class to Record

wkolowski:
This was already tried in #202, but I'm opening a separate PR to try it from scratch using a possibly slightly different approach.

PR #216, opened 2023-04-21
Make arguments of `state` and `label` explicit

wkolowski:
This PR is the first step towards #215.

I changed `state` and `label` from `VLSMType` to take the argument explicitly, i.e. now you have to write `state X` and `label X`.

Then I replaced occurrences of `state` and `label` with `state _` and `label _` using this script:

```bash
sed -i 's/@state _/state/g' $(find -name "*.v")

sed -i 's/ state)/ state _)/g' $(find -name "*.v")
sed -i 's/ state]/ state _]/g' $(find -name "*.v")
sed -i 's/ state}/ state _}/g' $(find -name "*.v")

sed -i 's/ state,/ state _,/g' $(find -name "*.v")
sed -i 's/ state;/ state _;/g' $(find -name "*.v")
sed -i 's/ state :=$/ state _ :=/g' $(find -name "*.v")
sed -i 's/ state$/ state _/g' $(find -name "*.v")

sed -i 's/ state ->/ state _ ->/g' $(find -name "*.v")
sed -i 's/ state \*/ state _ */g' $(find -name "*.v")

sed -i 's/@label _/label/g' $(find -name "*.v")

sed -i 's/ label)/ label _)/g' $(find -name "*.v")
sed -i 's/ label]/ label _]/g' $(find -name "*.v")
sed -i 's/ label}/ label _}/g' $(find -name "*.v")

sed -i 's/ label,/ label _,/g' $(find -name "*.v")
sed -i 's/ label;/ label _;/g' $(find -name "*.v")
sed -i 's/ label :=$/ label _ :=/g' $(find -name "*.v")
sed -i 's/ label$/ label _/g' $(find -name "*.v")

sed -i 's/ label ->/ label _ ->/g' $(find -name "*.v")
sed -i 's/ label \*/ label _ */g' $(find -name "*.v")
```

I filtered out the comments that were affected by this using `git add -p`, so that it only changes code.

I manually fixed the failing modules (there were only three) and manually fixed failing modules in casper-cbc-proofs.

What remains to be done is to fill the underscores with actual arguments so it reads better, but besides that the PR is ready for review. The best way to review is probably commit-by-commit.


theories/VLSM/Core/VLSM.v
```diff
@@ -391,8 +394,8 @@ Context
 
 Definition type := vtype vlsm.
 Definition machine := vmachine vlsm.
-Definition vstate := @state _ type.
-Definition vlabel := @label _ type.
+Definition vstate := state type.
+Definition vlabel := label type.
```
bmmoore, line 398: Now that `state` and `label` take an explicit VLSM argument I think `vstate` and `vlabel` should be remove.
As a preliminary test, you could see where changing them to `Notation`s breaks code that tries to do some unfolding or reduction on `vstate` and `vlabel`.

bmmoore (changes requested):
I think we need to be sure that the `_` can be filled in nicely before merging this.  (The code changes look fine in themselves).

If you could use a `VLSM` name as in `state X` for some VLSM `X` I would be happy (but we should also eliminate `vstate` in favor of `state` then), but at the moment  I think the argument must be a `VLSMType` expression as in `state (vtype X)`.

This can probably be allowed by making `vtype` a Coercion.

For the next step of filling in the `_`, I think it should be acceptable to sometimes leave the `_`, if it's not very illuminating. And, this AWK script might suggest reasonable names:
```awk
BEGIN { last_var="_" }
/[[({][^ ]+ : VLSM / { match($0,"[([{]([^ ]+) : VLSM ",parts); last_var=parts[1]; }
// { gsub(/\<state _\>/,"state " last_var); print}
```

wkolowski:
@bmmoore I filled in the `_` (except a few which I though were not very helpful).

I will leave adding the coercion for a separate PR, as it poses some problems (it makes a few things break). The same for `vstate` - I tried doing it right away, but it should be done in a separate step.

wkolowski:
@bmmoore I changed `state (type ...)` and `label (type ...)` to use `vstate` and `vlabel` instead, so that the style is consistent across the codebase after this PR is merged. I will remove `vstate` and `vlabel` in the next PR.

bmmoore (approved):
This is fine as an intermediate step, if you're still planning to eliminate `vstate` and `vlabel` in further work.

wkolowski (commented):
Pre-commit self-review.

PR #217, opened 2023-04-25
Replace `type` with `vtype` everywhere and delete `type`

wkolowski:
Turning `vtype` into a coercion is a little tricky, so I decided to get rid of `type` first.


wkolowski (commented):
Self-review.

traiansf (approved):

PR #218, opened 2023-04-26
Make `vtype` a coercion

wkolowski:
Besides making `vtype` a coercion, I made sure that this coercion is actually used everywhere (and `vtype` is explicitly used nowhere) and also made sure that variables named `vtype` were renamed to `T` to make things cleaner.

The script used to substitute `(vtype whatever)` -> `whatever`:

```bash
FILES=$(grep -l "vtype" $(find -name "*.v"))

sed -Ei "s/\(vtype ([a-zA-Z0-9'_]*)\)/\1/g" $FILES
sed -i 's/(vtype (IM i))/(IM i)/g' $FILES
sed -i 's/(vtype (IM j))/(IM j)/g' $FILES
```

### DIFF
--- a/theories/VLSM/Core/AnnotatedVLSM.v
+++ b/theories/VLSM/Core/AnnotatedVLSM.v
@@ -96,11 +96,11 @@ Lemma annotate_trace_item_project
   (k : annotated_state -> list (@transition_item _ annotated_type))
   (sa : annotated_state)
   : pre_VLSM_embedding_finite_trace_project
-      annotated_type (vtype X) id original_state
+      annotated_type X id original_state
       (annotate_trace_item item k sa)
       = item ::
         pre_VLSM_embedding_finite_trace_project
-            annotated_type (vtype X) id original_state
+            annotated_type X id original_state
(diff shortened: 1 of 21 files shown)


wkolowski (commented):
Self-review.

traiansf (approved):

bmmoore (commented):
I don't see why this should try to eliminate every use of `vtype`. Sometimes it could be helpful to be explicit about getting the type. Also, in the several cases where the arguments had `(vtype X) (vmachine X)`, it's strange to change to `X (vmachine X)`, rather than trying to eliminate the redundancy and need just one `X` argument, or leaving the original correspondence between the arguments. It could be confusing if `vtype` is written and then not printed out. If you want to make it more like an "only parsting" notation try `Set Printing Coercion vtype.`

traiansf:
My intuition was that you would still be able to use `type X` instead, if needed. And i thought that eventually, `X` and `machine X` would be merged in a single argument, but it's being done in smaller steps. Is that correct @wkolowski ?

PR #220, opened 2023-04-27
Make `vmachine` a coercion

wkolowski:
This PR is analogous to #218, but there are a few differences:
- I left explicit uses of `vmachine` in the notations `VLSM_incl` and `VLSM_eq`
- I had to fix some proofs here and there.


traiansf (approved):
I have to agree with Brandon, though, that it looks strange to have things like 
`apply (@VLSM_incl_input_valid_transition _ Mi Mi RMi)`

PR #221, opened 2023-04-27
Remove uses of Existing Instance from VLSM.v

wkolowski:
The use of `Existing Instance` for `TypeX` and `MachineX`, two aliases defined inside a section in VLSM.v has been bothering me for some time - now it can be easily removed.


traiansf (approved):
Looks fine

PR #223, opened 2023-05-04
Remove notations from Validator.v

wkolowski:
Now that arguments of `state` and `label` are explicit, notations in Validators.v are not needed anymore.


bmmoore (approved):
Looks good

PR #233, opened 2023-05-11
Change `VLSMMachine` from Class to Record

wkolowski:


bmmoore (approved):
Looks good.

I would like to avoid `Arguments ... : rename.`, but I don't know any other way to name that argument.

PR #236, opened 2023-05-23
Refactor implicit arguments

wkolowski:
In the previous refactoring a pattern emerged where we often set implicit arguments as `Arguments name {message vlsm} ..., {message} vlsm ...` so that we can either explicitly give the vlsm or not, depending on whether it can be inferred from the later arguments or not.

In this PR I applied this pattern to `state` and `label` (for which I missed it last time) as well as for some autogenerated constructors named `Build_...` and for `option_initial_message_prop`.

I also removed some uses of `@` where it was not needed.

### DIFF
--- a/theories/VLSM/Core/VLSM.v
+++ b/theories/VLSM/Core/VLSM.v
@@ -26,8 +26,21 @@ Record VLSMType (message : Type) : Type :=
   label : Type;
 }.
 
-Arguments state {_} _, {_ _}.
-Arguments label {_} _, {_ _}.
+(*
+  We set implicit arguments for [state] in such a way that we can provide the
+  argument either explicitly as <<state X>> or we can leave it out, writing
+  just <<state>>, allowing Coq to infer it. We do the same for [label].
+*)
+Arguments state {message v}, {message} v.
+Arguments label {message v}, {message} v.
+
+(*
+  Unfortunately, with the above implicit argument settings the argument is not
+  displayed (e.g. in the context during a proof it's displayed as <<state>>), so
+  we force Coq to always display this argument using an "only printing" notation.
+*)
+Notation "'state' X" := (@state _ X) (at level 10, only printing).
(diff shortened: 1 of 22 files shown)


traiansf (approved):

theories/VLSM/Core/VLSM.v
```diff
@@ -26,8 +26,21 @@ Record VLSMType (message : Type) : Type :=
...
+Arguments state {message v}, {message} v.
+Arguments label {message v}, {message} v.
+
+(*
+  Unfortunately, with the above implicit argument settings the argument is not
+  displayed (e.g. in the context during a proof it's displayed as <<state>>), so
+  we force Coq to always display this argument using an "only printing" notation.
+*)
+Notation "'state' X" := (@state _ X) (at level 10, only printing).
+Notation "'label' X" := (@label _ X) (at level 10, only printing).
```
wkolowski, line 43: Take a look at the new settings.

bmmoore: A big thing I missed, this makes `state` a keyword. So we absolutely shouldn't have this on by default. Other people trying to use a library probably want to use `state` as a name in their existing code.

Another thing any Notation would get wrong is quantified names. If another `state` is in scope so our `state` has to be written as `VLSM.state`, the notation would still print as `state`.

To generally avoid confusion it might be better to make it obviously a Notation by using some symbols, and we also wouldn't need `only printing`. Following the convention from some operators, `state@{X}` seems like reasonable syntax for specifying the type. That way anyone wondering where it comes from might try to locate it as a notation, rather than getting the mysterious parse error:

```
Toplevel input, characters 22-27:
> Locate state .
>        ^^^^^
Error: Syntax error: [locatable] expected after 'Locate' (in [command]).
```

wkolowski: I decided to revert this change. Better to just keep writing `state _` in a few places rather than come up with more layers of notations.

PR #239, opened 2023-06-07
upgrade to Dune Coq 0.6 and add Coq 8.18 to CI

palmskog:
This is an experimental PR to evaluate the feasibility of the new 3.8 series of releases of Dune with support for Coq composition with installed libraries. We are waiting for some Dune bugs to be fixed and for Docker images to be upgraded before this can go out of draft, notably ocaml/dune#7893 and coq-community/docker-coq#56

### DIFF
--- a/theories/VLSM/Core/VLSM.v
+++ b/theories/VLSM/Core/VLSM.v
@@ -2804,6 +2804,42 @@ Proof.
   - by apply finite_valid_trace_init_to_alt_right_impl.
 Qed.
 
+Definition finite_valid_trace_from_to_alt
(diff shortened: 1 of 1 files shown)


palmskog:
We are now waiting for https://github.com/coq-community/docker-coq/issues/57

palmskog:
This PR has been repurposed to switch to Dune-Coq 0.6, since 0.8 has problems with transitive dependencies. In any case we preserve coq_makefile builds, so Dune is only used in CI.

We now introduce Coq 8.18 into CI.

traiansf (approved):

PR #240, opened 2023-06-08
New coding convention: use lemma/theorem and tactics when proving propositions

wkolowski:
I changed proofs that were using `Definition` to `Lemma`, and the proofs use the `exact` tactic now.

This PR also contains some other changes, like added type annotations, etc.

### DIFF
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -128,6 +128,27 @@ Definition lv_message_observations (s : state) (target : index) : set lv_event :
   set_union (lv_sent_observations s target) (lv_received_observations s target).
 ```
 
+- do not use definitions to prove propositions; use lemma/theorem and tactics instead
+
+Not recommended:
+```coq
+Definition finite_trace_partial_map_app
+  : forall l1 l2, finite_trace_partial_map (l1 ++ l2) =
+    finite_trace_partial_map l1 ++ finite_trace_partial_map l2
+  := map_option_app _.
+```
+
+Recommended:
+```coq
+Lemma finite_trace_partial_map_app :
+  forall l1 l2 : list (transition_item TX),
+    finite_trace_partial_map (l1 ++ l2) =
+    finite_trace_partial_map l1 ++ finite_trace_partial_map l2.
+Proof.
+  by apply map_option_app.
+Qed.
+```
+
 ### Theorems and lemmas
 
 - C-style name
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -128,7 +128,7 @@ Definition lv_message_observations (s : state) (target : index) : set lv_event :
   set_union (lv_sent_observations s target) (lv_received_observations s target).
 ```
 
-- do not use definitions to prove propositions; use lemma/theorem and tactics instead
+- When giving an explicit proof term for things that live in `Prop`, avoid using `Definition` with `:=`, because it is unclear whether the point is just to give the proof term or whether transparency also matters. Instead use `Lemma` and provide the proof term using the `exact` tactic, while marking transparency with `Qed` or `Defined`.
 
 Not recommended:
 ```coq
@@ -145,7 +145,7 @@ Lemma finite_trace_partial_map_app :
     finite_trace_partial_map (l1 ++ l2) =
     finite_trace_partial_map l1 ++ finite_trace_partial_map l2.
 Proof.
-  by apply map_option_app.
+  exact (map_option_app _).
 Qed.
 ```
 
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -128,7 +128,7 @@ Definition lv_message_observations (s : state) (target : index) : set lv_event :
   set_union (lv_sent_observations s target) (lv_received_observations s target).
 ```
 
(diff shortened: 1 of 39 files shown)


CODING_CONVENTIONS.md
```diff
@@ -128,6 +128,27 @@ Definition lv_message_observations (s : state) (target : index) : set lv_event :
   set_union (lv_sent_observations s target) (lv_received_observations s target).
 ```
 
+- do not use definitions to prove propositions; use lemma/theorem and tactics instead
```
palmskog, line 131: I think this is quite far in precision and accuracy from what @bmmoore and I discussed elsewhere. Specifically:
- we don't really talk about "propositions", but about terms that live in the sort `Prop` (which have Coq's weak form of proof irrelevance)
- we don't necessarily want people to use tactics beyond a single call to `exact`, and this is only to get the constants to become opaque

So, for example, one would write the "recommended" version as:
```coq
Proof. exact (map_option_app _). Qed.
```

wkolowski: By "proposition" I meant a thing living in `Prop`, I can reword it if you want.

I don't understand your second point. We don't want people to use `exact`, do we? We don't use it when a subgoal can be proved by giving a term, so why should we if the entire lemma can be proved by giving a term? Also, what's the point of writing the term explicitly if we don't want to rely on the exact term and make it opaque?

palmskog: The point is that writing explicit terms is the most stable and reliable way to write Coq constants. `exact` is precisely the right tactic to do this. Using tactics besides `exact` when they aren't needed leads to much harder-to-maintain code. Our goal here is to enable having opaque constant bodies even when using `Lemma`, and `Proof. exact <term>. Qed.` is the best way to accomplish this. I will request changes to any PR that changes the body of an existing `Definition` living in `Prop` to anything more loose than `exact`.

bmmoore:
```suggestion
- To give a proof term for something in `Prop`, prefer a `Proof` with `explicit` over a `Definition` body (with `:=`). because it can be unclear whether the goal was just to give an explicit proof term, or if transparency was also intentional. To give an explicit term instead use a `Proof` with `exact` use because it is not clear whether transparency is intended or a side effect
```

bmmoore (commented):
I also think the policy description should be clarified a bit, as @palmskog says.

The code changes look fine.

I don't mind the keyword changes in this PR, but it should be clear that it's not strictly against this policy to have a `Definition ... . Proof. by (exact ...). Qed.`

palmskog (changes requested):
This PR does too many things at once:

1. It introduces and describes new coding conventions that need to be carefully formulated and agreed on.
2. It changes a lot of existing terms to use tactic-based proofs.
3. It changes opacity (`Defined`/`Qed`) of many constants.

I think all these three items should be split into different PRs and agreed on separately. To me, the question of opacity is nearly orthogonal to using `Definition` or `Lemma`. A rule of thumb is that one should have good reasons to change opacity of existing constants.

The fact that Coq has no good mechanism to separate opacity from the mode of defining a proof script (`Qed` vs. `Defined` vs. `X := y`) is for historical reasons more than anything else. We may get an attribute like `#[opacity=transparent]` in the near future.

wkolowski:
@bmmoore I think with the new wording `Definition ... Qed` is not recommended for `Prop` (we prefer `Lemma ... Qed`) but not strictly forbidden, whereas if it's something in `Type`, then the convention does not apply.

@palmskog I removed code changes from this PR and left only the update to coding conventions. I will open a separate PR with the code changes (`Definition` to `Lemma`) and then yet another one which changes `Defined` to `Qed` (as I think there is not reason for these proofs to be transparent).

CODING_CONVENTIONS.md
```diff
@@ -128,6 +128,27 @@ Definition lv_message_observations (s : state) (target : index) : set lv_event :
   set_union (lv_sent_observations s target) (lv_received_observations s target).
 ```
 
+- When giving an explicit proof term for things that live in `Prop`, avoid using `Definition` with `:=`, because it is unclear whether the point is just to give the proof term or whether transparency also matters. Instead use `Lemma` and provide the proof term using the `exact` tactic, while marking transparency with `Qed` or `Defined`.
```
palmskog, line 131:
```suggestion
- When explicitly defining a constant that lives in `Prop`, avoid using `Definition` with `:=`. Instead use `Lemma` (or `Theorem`, `Proposition`, etc.) and provide the constant term body using the `exact` tactic, while indicating the opacity with `Qed` or `Defined`.
```

CODING_CONVENTIONS.md
```diff
@@ -128,6 +128,27 @@ Definition lv_message_observations (s : state) (target : index) : set lv_event :
...
+
+Recommended:
+```coq
+Lemma finite_trace_partial_map_app :
+  forall l1 l2 : list (transition_item TX),
+    finite_trace_partial_map (l1 ++ l2) =
+    finite_trace_partial_map l1 ++ finite_trace_partial_map l2.
+Proof.
+  exact (map_option_app _).
+Qed.
```
palmskog, line 149:
```suggestion
Proof. exact (map_option_app _). Qed.
```

palmskog (approved):
This is fine by me now, but please consider the suggested reformulations.

CODING_CONVENTIONS.md
wkolowski: Changed.

CODING_CONVENTIONS.md
wkolowski: Changed.

PR #243, opened 2023-06-15
Change Definition to Lemma for proofs

wkolowski:


theories/VLSM/Core/Equivocators/MessageProperties.v
```diff
@@ -426,7 +426,8 @@ Context
   We define [has_been_received] for the [equivocator_vlsm] as being received by any
   of the internal machines.
 *)
-Definition equivocator_has_been_received  := equivocator_oracle (has_been_received X).
+Definition equivocator_has_been_received : state equivocator_vlsm -> message -> Prop :=
```
palmskog, line 429: The title of the PR is about definitions/lemmas living in `Prop`, which this constant doesn't. What's the motivation for doing these explicit type annotations? I don't think they are very helpful, and in particular, I don't think we should be going through existing code and adding them. Types of constants are already available as part of the coqdoc documentation and through `Check`.

theories/VLSM/Core/TraceableVLSM.v
```diff
@@ -195,7 +195,7 @@ Definition composite_state_destructor (s' : composite_state IM) (i : index)
   : list (composite_transition_item IM * composite_state IM) :=
   map (lift_to_composite_transition_item_state s' i) (state_destructor i (s' i)).
 
-Definition composite_state_size (s : composite_state IM) :=
+Definition composite_state_size (s : composite_state IM) : nat :=
```
palmskog, line 198: This is another unmotivated change.

palmskog (changes requested):
Please change the PR so it conforms to the title.

theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
```diff
@@ -1346,30 +1346,35 @@ Proof.
...
       equivocators_trace_project final_descriptors trX = Some (trY, initial_descriptors) /\
-      equivocators_state_project initial_descriptors sX = sY)
-  : partial_trace_project (sX, trX) = Some (sY, trY)
-  := proj2 (equivocators_partial_trace_project_characterization final_descriptors sX trX sY trY) H.
+      equivocators_state_project initial_descriptors sX = sY) :
+    partial_trace_project (sX, trX) = Some (sY, trY).
+Proof.
+  exact (proj2 (equivocators_partial_trace_project_characterization
+    final_descriptors sX trX sY trY) Hed).
+Defined.
```
wkolowski, line 1377: @palmskog Funny observation: at this point `H` is used for an instance of `Finite`. When the above thing was a `Definition`, using `H` for the hypothesis name was fine, but when it's a `Lemma`, Coq rejects it and says that `H` is already used. Is this kind of behaviour intended or is it a bug?

theories/VLSM/Core/Equivocators/MessageProperties.v
wkolowski: It seems I forgot to filter them out when I was splitting the PR into 3 parts. They are gone now.

As for why I always try to add them when they're missing, I find the lack of explicit type annotations very confusing and I do think they should be mandatory (and we should go over all existing code and add them).

To me having explicit annotations makes reading code more efficient, because when I need the type, I can just read it. If I don't intimately know the code, running the typechecker in my head may not be a viable option (and for many modules in VLSM it still is). Of course I can use `Check`, but it takes a couple seconds (and maybe up to 15-30 if I haven't checked the file up to that point, and maybe even a few minutes if I am looking at a branch that I didn't build yet). As for coqdoc, maybe I'm missing something, but I don't see the types anywhere if they were not explicitly written: https://runtimeverification.github.io/vlsm-docs/latest/coqdoc/VLSM.Core.Equivocators.MessageProperties.html#equivocator_has_been_received. But even if coqdoc had the types, switching between the editor and the browser takes some time, and looking up doc pages corresponding to the current file also takes some time (especially that the docs are named after the `(* *` header instead of the file name).

I just can't see why anyone wouldn't want to have explicit type annotations.

theories/VLSM/Core/TraceableVLSM.v
wkolowski: Fixed.

wkolowski:
@palmskog I have removed the unrelated changes, so you can re-review.

theories/VLSM/Core/Equivocators/MessageProperties.v
wkolowski: @traiansf @bmmoore What's your opinion on explicit type annotations?

theories/VLSM/Core/VLSMProjections/VLSMInclusion.v
```diff
@@ -95,12 +95,14 @@ Proof.
...
-  {MX MY : VLSMMachine T}
-  (X := mk_vlsm MX) (Y := mk_vlsm MY)
-  (Hincl : VLSM_incl X Y)
-  : VLSM_embedding X Y id id
-  := proj1 (VLSM_incl_embedding_iff MX MY) Hincl.
+Lemma VLSM_incl_is_embedding :
+  forall
+    {MX MY : VLSMMachine T}
+    (X := mk_vlsm MX) (Y := mk_vlsm MY),
+      VLSM_incl X Y -> VLSM_embedding X Y id id.
```
bmmoore, line 102: This layout is pretty awkward. I would keep some arguments before the `:`
(as seen on the very next lemma `VLSM_incl_is_embedding_finite_trace_project`)

```suggestion
Lemma VLSM_incl_is_embedding
  {MX MY : VLSMMachine T}
  (X := mk_vlsm MX) (Y := mk_vlsm MY) :
    VLSM_incl X Y ->
    VLSM_embedding X Y id id.
```

theories/VLSM/Lib/ListExtras.v
```diff
@@ -152,11 +152,10 @@ Proof.
   by split; [| apply Hsub].
 Qed.
 
-Definition app_cons {A}
-  (a : A)
-  (l : list A)
-  : [a] ++ l = a :: l
-  := eq_refl.
+Lemma app_cons :
+  forall {A : Type} (a : A) (l : list A),
```
bmmoore, line 156: For consistency with the other definitions around here, keep the `{A}` before `:`. Also, I think we can leave out the ` : Type` on obvious type variables, even if we generally want to annotate things.
```suggestion
Lemma app_cons {A} :
  forall (a : A) (l : list A),
```

bmmoore (approved):
The actual Definition body to Proof changes are fine.

Reformatting type declarations to follow current conventions along with changing the rest of the definition is good, but some changes seem inconsistent, or unnecessarily changing away from the previous style (and the style of nearby definitions in the file) in ways that are not suggested by the coding conventions, nor big improvements in readability.

theories/VLSM/Core/Equivocation.v
```diff
@@ -547,17 +547,23 @@ Class HasBeenSentCapability : Type :=
...
   negate_oracle has_been_sent.
 
-Definition has_been_sent_no_inits `{HasBeenSentCapability} :
+Lemma has_been_sent_no_inits `{HasBeenSentCapability} :
   forall s : state vlsm,
-    initial_state_prop vlsm s -> ∀ m : message, ~ has_been_sent s m
-  := oracle_no_inits _ _ (has_been_sent_stepwise_props).
+    initial_state_prop vlsm s ->
+  forall m : message,
+    ~ has_been_sent s m.
```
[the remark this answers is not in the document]
palmskog: I agree with @bmmoore that the old formatting was OK, but a minor issue.

theories/VLSM/Core/Equivocation.v
```diff
@@ -547,17 +547,23 @@ Class HasBeenSentCapability : Type :=
...
+Defined.
 
-Definition has_been_sent_step_update `{HasBeenSentCapability} :
+Lemma has_been_sent_step_update `{HasBeenSentCapability} :
   forall (l : label _) (s : state _) (im : option message) (s' : state _) (om : option message),
     input_valid_transition (pre_loaded_with_all_messages_vlsm vlsm) l (s, im) (s', om) ->
-  forall msg,
-    has_been_sent s' msg <-> (om = Some msg \/ has_been_sent s msg)
-  := oracle_step_update _ _ has_been_sent_stepwise_props.
+  forall msg : message,
```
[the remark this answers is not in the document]
palmskog: In this particular case the type is obvious, so probably it should be omitted. But minor issue.

theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
palmskog: I don't think it's a bug, it's just that Coq's variable naming algorithm takes into account the form of the declaration (`Lemma` or `Definition`). The best practice is to not rely on automatic names at all, but difficult in practice.

palmskog (approved):
This looks good to me now, but please take Brandon's comments into account before merging.

theories/VLSM/Core/Equivocation.v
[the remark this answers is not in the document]
wkolowski: I changed it back (except that I also changed `∀` to `forall`).

theories/VLSM/Core/Equivocation.v
```diff
@@ -720,17 +726,22 @@ Class HasBeenReceivedCapability : Type :=
 Definition has_not_been_received `{HasBeenReceivedCapability} : state_message_oracle :=
   negate_oracle has_been_received.
 
-Definition has_been_received_no_inits `{HasBeenReceivedCapability} :
+Lemma has_been_received_no_inits `{HasBeenReceivedCapability} :
   forall s : state vlsm,
-    initial_state_prop vlsm s -> ∀ m : message, ~ has_been_received s m
-  := oracle_no_inits _ _ has_been_received_stepwise_props.
+    initial_state_prop vlsm s ->
+  forall m : message, ~ has_been_received s m.
```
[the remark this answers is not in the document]
wkolowski: Changed (except `∀` is now `forall`).

theories/VLSM/Core/Equivocation.v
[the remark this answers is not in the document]
wkolowski: I'll keep it for now. As for the conventions, for me there are barely any type annotations that are unnecessary. I started a conversation about this topic as a reply to one of Karl's suggestions, you might want to have a say there.

theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
```diff
@@ -1346,30 +1346,35 @@ Proof.
...
-  (Hpr_tr : partial_trace_project (sX, trX) = Some (sY, trY))
-  : not_equivocating_equivocator_descriptors IM final_descriptors (finite_trace_last sX trX) /\
-    exists initial_descriptors,
-      equivocators_trace_project final_descriptors trX = Some (trY, initial_descriptors) /\
-      equivocators_state_project initial_descriptors sX = sY
-  := proj1 (equivocators_partial_trace_project_characterization
-      final_descriptors sX trX sY trY) Hpr_tr.
+  (Hpr_tr : partial_trace_project (sX, trX) = Some (sY, trY)) :
+    not_equivocating_equivocator_descriptors IM final_descriptors (finite_trace_last sX trX) /\
+      exists initial_descriptors,
```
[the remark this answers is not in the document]
wkolowski: It's just inconsistency. I only add the annotations if I notice they aren't there :)

theories/VLSM/Lib/ListExtras.v
wkolowski: Changed.

theories/VLSM/Core/Equivocation.v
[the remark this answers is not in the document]
palmskog: I completely disagree with a convention saying anything to the effect of "bare any type annotations are unnecessary". There is a long tradition in Standard ML, OCaml, HOL and Coq of using type annotations to *control* the automatic type inference. Hence, in many Coq and OCaml developments, when a type annotation is seen, it's an indication that the type inference was "wrong" about something (e.g., too general type)

PR #244, opened 2023-06-15
Change `Defined` to `Qed` for proofs

wkolowski:
I believe that the fact that these proofs were originally transparent was an entirely accidental side effect of using `Definition`. Since nothing depends on their being transparent and nothing breaks in neither repo, I think it's ok to make them opaque.

This PR touches only the proofs that were recently changed from `Definition` to `Lemma`.


bmmoore:
That argument sounds reasonable to me, but currently the diff is poor and includes many Definition to Lemma changes. I think you need to rebase this branch on top of the rebased version of `change-definition-to-lemma`.

palmskog:
@wkolowski please rebase this, and I will review it afterwards. You may want to merge #243 first.

PR #246, opened 2023-06-20
Refactor preloaded vlsm

wkolowski:
Summary of changes:
- Redefine `pre_loaded_with_all_messages_vlsm` to be `pre_loaded_vlsm (fun _ => True)`.
- Move these definitions further towards the end of Composition.v
- Fix the proofs that were broken (often this means doing an additional `cbn; right`)
- Remove some lemmas about `pre_loaded_with_all_messages_vlsm` which are now subsumed by the analogous lemmas for `pre_loaded_vlsm`.
- Refactor some proofs to get rid of uses of these removed lemmas.

Thanks to this refactoring I was able to get rid of some of the ugliest proofs in the codebase (these using lots of Ltac `match`es). The cost is that a few proofs got a bit more complicated, but on net I think it's a huge simplification.

As a bonus, this refactoring allowed me to spot some places which were inadvertently using doubly-preloaded vlsms. These are the parity example and Paxos. I will fix these in the update dependency PR.

### DIFF
--- a/theories/VLSM/Core/VLSM.v
+++ b/theories/VLSM/Core/VLSM.v
@@ -73,18 +73,6 @@ Definition option_initial_message_prop
   {message : Type} {T : VLSMType message} (M : VLSMMachine T)
   : option message -> Prop := from_option (@initial_message_prop _ _ M) True.
 
-Definition VLSMMachine_pre_loaded_with_messages
-  {message : Type} {T : VLSMType message} (M : VLSMMachine T)
-  (initial : message -> Prop)
-  : VLSMMachine T
-  :=
-  {| initial_state_prop := @initial_state_prop _ _ M
-  ; initial_message_prop := fun m => @initial_message_prop _ _ M  m \/ initial m
-  ; s0 := @s0 _ _ M
-  ; transition := @transition _ _ M
-  ; valid := @valid _ _ M
-  |}.
-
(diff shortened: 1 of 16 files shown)


theories/VLSM/Core/VLSM.v
```diff
@@ -728,8 +708,8 @@ Qed.
 
 (**
   For VLSMs initialized with many initial messages such as
-  the [composite_vlsm_induced_projection] or the [pre_loaded_with_all_messages_vlsm],
-  the question of whether a [VLSM] [can_emit] a message <<m>> becomes more
+  the [composite_vlsm_induced_projection] the question of
```
bmmoore, line 711: This should probably still mention `[pre_loaded_vlsm]` because it is much more common than a `composite_vlsm_induced_projection`.

bmmoore (approved):
Looks pretty good to me.
One suggestion on a comment.

Also, I don't think proofs should expand be more than a line or two for a single affected point, so the changes in Equivocation.v are surprising. Could you avoid that with an additional lemma or two about these preloaded vlsms?

theories/VLSM/Core/VLSM.v
wkolowski: I deleted it because preloaded vlsms haven't been defined yet at that point, but I brought it back after your suggestion.

PR #250, opened 2023-06-23
Refactor duplicated free composition lemmas in Composition.v

wkolowski:
For some of the duplicated proofs in Composition.v that were introduced with the new definition of free composition, I was able to derive the version for `composite_vlsm` from the one for `free_composite_vlsm`, possibly simplifying the proof a bit in the process. I was also able to bring back the old proof of the lemma `same_IM_preloaded_free_embedding`, which was derived from an analogous proof for the constrained composition.

Some lemmas (those that are named like `...relevant_components...` are not possible to refactor without making bigger changes and will be dealt with in a future PR. Also, some duplicated lemmas remain in Equivocation.v and will also be dealt with in a future PR.

### DIFF
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -880,25 +880,6 @@ Ltac state_update_simpl :=
   precise.
 *)
 
-Lemma valid_state_project_preloaded_to_preloaded
-      message `{EqDecision index} (IM : index -> VLSM message) constraint
-      (X := composite_vlsm IM constraint)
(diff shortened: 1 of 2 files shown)


traiansf (approved):
Looks reasonable to me; however, the fact that you did both formatting and actual code changing, made it harder to review.

wkolowski:
@traiansf The free lemmas were located after the corresponding constrained lemmas. To derive the latter from the former, I had to switch their places, which damaged the diff readability pretty badly, so I was less on guard for reviewability of the PR. Also probably it has something to do with the fact that Karl doesn't need to review this :)

PR #251, opened 2023-06-27
Remove the lemma `related_components` and some lemmas it depends on

wkolowski:
The removed lemmas were duplicated when introducing new definition of `free_composite_vlsm`.

This PR requires a fix in casper-cbc-proofs, which I already have.

### DIFF
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -1247,19 +1247,9 @@ Lemma irrelevant_components_one
   let res := snd (composite_apply_plan IM s [ai]) in
   (res i) = (s i).
 Proof.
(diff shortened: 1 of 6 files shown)


wkolowski:
@traiansf Looks like a tiny refactoring PR #252 that was based off this branch was merged into it rather than into master... you can ignore all the changes which are outside Composition.v as they were already reviewed by Karl.

traiansf (commented):
I could approve, but I don't understand why you need to remove that lemma; I understand that it's not currently being used (I assume you have checked the casper-cbc repo --- this lemma was written around the work on the list validator), but the lemmas are saying something interesting about the ability to permute actions.
Hence, I would like to hear more arguments for their removal before approving.

traiansf (approved):

PR #252, opened 2023-06-27
Remove many uses of `match goal`

wkolowski:


theories/VLSM/Lib/TopSort.v
```diff
@@ -190,8 +190,7 @@ Proof.
   - left.
     specialize (Forall_forall P l0); intros [Hall _].
     specialize (Hall HPl0 x Hin).
-    match goal with |- ?X = 0  => cut (X <= 0) end.
-    lia.
+    cut (length (filter (fun b => precedes b a) l0) <= 0); [by lia |].
```
palmskog, line 193: The others are fine now, but I don't see much of a gain with this one. The consensus in the MathComp community is that mentioning as few constants in as possible in pattern matches is a good thing (see, e.g., [this](https://inria.hal.science/inria-00258384/)).
```suggestion
    match goal with |- ?X = 0  => cut (X <= 0) end; [by lia|].
```

palmskog (commented):
This looks mostly fine, but I disagree with the last one.

theories/VLSM/Lib/TopSort.v
wkolowski: Changed.

palmskog (approved):
Now this is a clear improvement, please merge as-is.

PR #254, opened 2023-07-04
Remove some uses of match type

wkolowski:


theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
```diff
@@ -351,13 +349,10 @@ Proof.
...
   ; intros
-  ; match type of Ht with
-    | (let (_, _) := ?t in _) = _ =>
-      destruct t as (si', om') eqn: Ht'
-    end
+  ; case_match
   ; inversion Ht; subst; clear Ht
   ; rewrite state_update_eq in Hchar
-  ; specialize (Hchar _ Hv Ht')
+  ; specialize (Hchar _ Hv H1)
```
palmskog, line 355: This trades an explicit name for a machine-generated name that wouldn't survive name changes/mangling. I think that's a strict negative even when the number of lines decreases.

palmskog (commented):
Most look reasonable, but please address the one I commented on if you want this approved.

theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
wkolowski: I changed it to `case_match eqn: Ht'` to preserve the name - I didn't know that you can use `eqn:` with `case_match`,

palmskog (approved):

PR #255, opened 2023-07-04
Refactor EquivocatorsCompositionProjections.v

wkolowski:
At the beginning of the module EquivocatorsCompositionProjections there was a big `Context` with abbreviations whose only purpose was to avoid writing one argument.
```
  (equivocator_descriptors := equivocator_descriptors IM)
  (equivocators_no_equivocations_vlsm := equivocators_no_equivocations_vlsm IM)
  (equivocators_state_project := equivocators_state_project IM)
  (equivocator_IM := equivocator_IM IM)
  (equivocator_descriptors_update := equivocator_descriptors_update IM)
  (proper_equivocator_descriptors := proper_equivocator_descriptors IM)
```
A very unpleasant side effect of these abbreviations was that in proof mode, they were cluttering the proof context, which decreased readability and necessitated lots of scrolling just to see what the goal is.

In this PR I removed these abbreviations by just writing out the `IM` argument in full (or sometimes, just `_`). The result is that the proof contexts are much shorter now and less scrolling is needed - overall, stepping over the proofs is much nicer now.

Some proofs required tiny fixes because a `Hint Unfold`, which was no longer necessary, was also removed.

### DIFF
--- a/theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
+++ b/theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
@@ -17,8 +17,7 @@ Context {message : Type}
   `{finite.Finite index}
   (IM : index -> VLSM message)
   `{forall i : index, HasBeenSentCapability (IM i)}
-  (equivocator_IM := equivocator_IM IM)
-  (FreeE := free_composite_vlsm equivocator_IM)
+  (FreeE := free_composite_vlsm (equivocator_IM IM))
   (PreFreeE := pre_loaded_with_all_messages_vlsm FreeE)
   (Free := free_composite_vlsm IM)
   (PreFree := pre_loaded_with_all_messages_vlsm Free)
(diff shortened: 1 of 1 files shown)


theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
```diff
@@ -1733,20 +1720,20 @@ Proof.
...
-  (is : composite_state sub_equivocator_IM)
-  (tr : list (composite_transition_item sub_equivocator_IM))
+  (is : composite_state _)
+  (tr : list (composite_transition_item _))
   (Htr : finite_valid_trace SeededXE is tr)
   (final_state := finite_trace_last is tr)
-  (final_descriptors : (equivocator_descriptors sub_IM))
-  (Hproper : proper_equivocator_descriptors sub_IM final_descriptors final_state)
+  (final_descriptors : equivocator_descriptors _)
+  (Hproper : proper_equivocator_descriptors _ final_descriptors final_state)
```
palmskog, line 1728: It's strange to me to use the wildcard as part of lemma statements. If the type is long to spell, why not use a `let ... in`? Or use some local (or even global) notation?

Sometimes, it's inevitable that Coq code can't be understood effectively without running Coq, but this kind of seemingly unnecessary wildcarding makes it difficult to read the code and understand it (guess what Coq infers).

traiansf (approved):
Please address @palmskog 's suggestions before merging

theories/VLSM/Core/Equivocators/EquivocatorsCompositionProjections.v
wkolowski: Fixed.

I replaced some of the `_` with inline terms and other with an abbreviation local to the theorem and thanks to this I was able to save even more screen space in proof mode.

PR #257, opened 2023-07-06
add Coq 8.17 to CI

palmskog:
I think it's a good idea as a general policy to support the latest two Coq versions with the project, as [Std++ does](https://gitlab.mpi-sws.org/iris/stdpp/#prerequisites). To this end, I add 8.17 to CI that must pass for a PR to be merged. Coq's compatibility policy ensures that this is always possible.

This means that we'd do a small "sliding window" of supported Coq versions going forward. For example, some time after Coq 8.18.0 is out, we switch to testing Coq 8.17 and 8.18.

### DIFF
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -14,6 +14,7 @@ jobs:
       matrix:
         image:
           - 'coqorg/coq:8.16-ocaml-4.14-flambda'
+          - 'coqorg/coq:8.17-ocaml-4.14-flambda'
       fail-fast: false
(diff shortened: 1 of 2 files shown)


traiansf (approved):

wkolowski (approved):

palmskog:
I will wait until week of July 10 to merge this, since we'd like Brandon to give thumbs-up as well.

palmskog:
Since we can reverse the policy of checking two versions later if inconvenient, I'm merging this now.

PR #258, opened 2023-07-07
document conventions for record names, including classes

palmskog:
We have agreed to start using `mk` prefix for record constructors, in particular to avoid our code being littered with autogenerated `Build_<RecordTypeName>`. This PR makes it official, and is the first step towards changing these annoying constructor names.

### DIFF
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -67,7 +67,8 @@ Section sec_ELMOComponent_lemmas.
 
 ### Type classes
 
-- CamelCase name
+- CamelCase for class (type) name
+- CamelCase for constructor name using prefix `mk` (when construction via `Instance` might not be sufficient)
 - field declaration with C-style naming on separate line, with 2 spaces of indentation
 - the `;` in the last field should not be omitted
 - it's recommended to include the sort annotation, especially when it's `Prop`
@@ -161,19 +162,19 @@ Lemma sync_some (s : vstate X) (from to : index) :
 
 ### Records
 
-- C-style name
-- CamelCase for constructor name
+- C-style for record (type) name
+- CamelCase for constructor name using prefix `mk`
 - field declaration with C-style naming on separate line, with 2 spaces of indentation
 - the `;` in the last field should not be omitted
 - it's recommended to include the sort annotation, especially when it's `Prop`
 
 Example:
 ```coq
-Record simp_lv_event : Type := SimpObs
+Record simp_lv_event : Type := mkSimpObs
 {
-  get_simp_event_type : simp_lv_event_type;
-  get_simp_event_subject : index;
-  get_simp_event_state : (@state index index_listing);
+  simp_lv_event_type : simp_lv_event_type;
+  simp_lv_event_subject : index;
+  simp_lv_event_state : (@state index index_listing);
 }.
 ```
 
--- a/CODING_CONVENTIONS.md
+++ b/CODING_CONVENTIONS.md
@@ -174,7 +174,7 @@ Record simp_lv_event : Type := mkSimpObs
 {
   simp_lv_event_type : simp_lv_event_type;
   simp_lv_event_subject : index;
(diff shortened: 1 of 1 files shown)


CODING_CONVENTIONS.md
```diff
@@ -161,19 +162,19 @@ Lemma sync_some (s : vstate X) (from to : index) :
...
 ```coq
-Record simp_lv_event : Type := SimpObs
+Record simp_lv_event : Type := mkSimpObs
 {
-  get_simp_event_type : simp_lv_event_type;
-  get_simp_event_subject : index;
-  get_simp_event_state : (@state index index_listing);
+  simp_lv_event_type : simp_lv_event_type;
+  simp_lv_event_subject : index;
+  simp_lv_event_state : (@state index index_listing);
```
wkolowski, line 177:
```suggestion
  simp_lv_event_state : @state index index_listing;
```

wkolowski (approved):
The idea is ok, but we should agree on the exact form of the prefix. Since constructor names are supposed to be `CamelCase`, I think the prefix should be `Mk` (starting with uppercase) rather than `mk` (lowercase). This is also what can be found in the codebase, with `Mk` much more frequent than `mk` (the only `mk`-prefixed constructor is `mk_vlsm`).

I think you should wait with merging till Monday, to let @bmmoore and @traiansf weigh in on this.

As a side note, the conventions on record naming seem not adhered to, with around 50% of records named in `CamelCase`, and another 50% following `snake_case`.

palmskog:
For the record, the `mk` prefix is overwhelmingly used by other Coq projects compared to `Mk`:
- https://sherlocoq.sirref.org/?q=mk (several hundred thousand hits)
- https://sherlocoq.sirref.org/?q=Mk (a few thousand hits)

bmmoore:
I think it would be quite rare to actually have name collisions between a record constructor and type constructor if we just used a `Mk` prefix, but avoiding it with `mk` is reasonable enough even if it is a bit odd looking. The original `Build_TypeName` form wasn't CamelCase nor snake_case, so there is some precedent for an unusual name format.

Why do only type classes have the "if expected to be used explicitly" clause?

It seems part of the point is to allow shorter constructor names than just mkTypeName, but we should also allow that (especially if it's required to always explicitly name record constructors), so we should specifically say you can't risk confusion with existing identifiers by using mkConName when there is some existing type or constructor ConName (other than the record you are defining).

Those searches are not even restricted to prefixes. " mk" vs. " Mk" and "(mk" vs "(Mk" are a bit better but don't restrict to constructor. Those still have about 10x more hits for "mk"-prefixed things so it's probably more popular also for constructors, but many uses seem like they are perhaps plain function definitions or lemmas rather than constructors.

palmskog:
> Why do only type classes have the "if expected to be used explicitly" clause?

The idea is that type classes that live in `Prop` are not meant to be constructed manually from scratch, but only through the typeclass inference machinery and via `Instance` declarations. In certain cases, we might observe that a `Prop` typeclass gets manually constructed a lot, and then we can give its constructor a manual name. At least that is the idea proposed here.

palmskog:
@traiansf any opinions here on `mk` vs `Mk` and related issues?

While `mk` is also commonly used for regular function names, I think this is mostly due to those functions operating on constructors.

traiansf:
While I don't care much for naming, I would incline towards `mk_` to be uniform. We already use `sec_` for sections regardless of camel case

palmskog:
I'm fine with `mk_` as a compromise, in analogy with `sec_`. @wkolowski is that OK with you?

wkolowski:
@palmskog I think how our records are named is also important for what the prefix should be. If we want the constructor name to be prefix + type name, then `mk_` make the most sense for `snake_case`-named records, whereas `Mk` makes the most sense for `CamelCase`-named records, but `mk` is also fine there.

@traiansf Section names are `snake_case` with `CamelCase` ocurring only because some things are named in this way.

I'd say, let's just put `mk` in the coding conventions and see where it leads us.

PR #265, opened 2023-07-17
Decouple `annotated_vlsm` from its constraint

wkolowski:
Currently `annotated_vlsm` modifies the validity constraint of the underlying VLSM. The goal of this PR is to change it, so that the new `annotated_vlsm` deals only with the annotation, whereas the constraining is done by `constrained_vlsm`.

### DIFF
--- a/theories/VLSM/Core/AnnotatedVLSM.v
+++ b/theories/VLSM/Core/AnnotatedVLSM.v
@@ -2,6 +2,7 @@ From VLSM.Lib Require Import Itauto.
 From stdpp Require Import prelude.
 From VLSM.Lib Require Import ListExtras.
 From VLSM.Core Require Import VLSM PreloadedVLSM VLSMProjections Validator Composition.
+From VLSM.Core Require Import ConstrainedVLSM.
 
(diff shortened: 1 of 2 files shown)


theories/VLSM/Core/AnnotatedVLSM.v
```diff
@@ -224,8 +222,9 @@ Context
     label (annotated_type Free annotation) ->
       annotated_state Free annotation * option message -> annotation)
   (AnnotatedFree : VLSM message :=
-    annotated_vlsm Free annotation initial_annotation_prop annotated_constraint
-      annotated_transition_state)
+    constrained_vlsm
```
traiansf, line 225: I wonder whether the AnnotatedFree should be constrained at all... It seems we could recover the results using a VLSM inclusion from constrained into free then the projection.

traiansf (approved):

theories/VLSM/Core/AnnotatedVLSM.v
wkolowski: I'll look into it.

PR #266, opened 2023-07-17
Use lemma `constrained_preloaded_incl` in a few places instead of reproving it inline

wkolowski:
There was a TODO mark in ConstrainedVLSM.v which suggested that the lemma `constrained_preloaded_incl` was being reproved inline in many places. I tried to hunt down these and use the lemma instead.


traiansf (approved):
It's fine, but I wouldn't have removed the extra lemmas, but instead made them definitions/instances of the existing proof.

PR #267, opened 2023-07-18
Updating the detailed building instructions

DafinaTrufas:
I've added a command to the detailed build guide, for installing the necessary C compiler, before creating the opam switch (Initially I got an error regarding this and this is the way I fixed it).


palmskog:
@DafinaTrufas I reorganized so that users are asked to install all opam requirements at once, and also added Fedora equivalent. I'll take the PR to merge, thanks for submitting.

palmskog (approved):

PR #271, opened 2023-07-18
Refactor many things related to `Context`s

wkolowski:
This PR follows #255, #268, #269, #270.

Summary of changes:
- Standardize indentation of `Context`s across the entire codebase.
- Remove abbreviations that were not used (I think this might be pretty important, because we gain clarity regarding what assumptions are needed for which theorems).
- Inline abbreviations in case where they were just cluttering the proof context in interactive mode.

### DIFF
--- a/theories/VLSM/Core/ByzantineTraces.v
+++ b/theories/VLSM/Core/ByzantineTraces.v
@@ -163,10 +163,10 @@ Qed.
 Section sec_pre_loaded_with_all_messages_byzantine_alt.
 
 Context
-    (PreLoaded := pre_loaded_with_all_messages_vlsm M)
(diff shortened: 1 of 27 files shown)


traiansf (approved):
Not sure this was really necessary, but if you insist...

wkolowski:
@traiansf I really hate it when I need to touch one of these proofs where the abbreviations eat up my entire screen. Also, it was really surprising to discover just how many of these were not used at all. For example, in LimitedByzantineTraces.v there was a `Context` which assumed

```
(Hsender_safety : sender_safety_alt_prop IM A sender :=
    channel_authentication_sender_safety _ _ _ Hchannel)
```

but didn't use it. You can think of this PR as a part of dealing with "where do we make which assumptions".

PR #272, opened 2023-07-20
Improve parity example coqdoc comments

DafinaTrufas:
In addition to adding more coqdoc comments (which I hope to be useful, maybe some of them are to obvious, I'm not sure), I made a few minor changes in some proofs, where I thought I found a shorter proof (for a utility lemma, using a stdpp lemma in the proof made it a bit shorter), or there was a redundant step (eg. a `cbn` which did nothing). Please let me know if that's ok.

### DIFF
--- a/theories/VLSM/Core/Examples/Parity.v
+++ b/theories/VLSM/Core/Examples/Parity.v
@@ -25,12 +25,13 @@ Proof. by intros n; split; intros [p Hp]; exists (-p); lia. Qed.
 
 (** Even right-hand side and difference implies even left-hand side 
     This lemma will be useful when proving the final result of this section, because of the way 
-    we defined the transitions in the Parity VLSM *)
+    we defined the transitions in the Parity VLSM
+*)
 Lemma Zeven_sub_preserve_parity :
   forall (n m : Z), Z.Even n -> Z.Even (m - n) -> Z.Even m.
 Proof. by intros m n [m'] [n']; exists (m' + n'); lia. Qed.
 
-(** Parity is preserved by taking sum *)
+(** Parity is preserved by addition *)
 Lemma Zeven_equiv_plus :
   forall (n m : Z), (Z.Even n <-> Z.Even m) -> Z.Even (m + n).
 Proof.
@@ -45,7 +46,7 @@ Proof.
     by rewrite Zeven_equiv, <- Hparity, <- Zeven_equiv.
 Qed.
 
-(** Parity is preserved by taking difference *)
+(** Parity is preserved by subtraction *)
 Lemma Zeven_equiv_minus :
   forall (n m : Z), (Z.Even n <-> Z.Even m) -> Z.Even (m - n).
 Proof.
@@ -57,7 +58,6 @@ Proof.
   by apply Zeven_unary_minus.
 Qed.
 
-(** Even sum term implies parity of the sum depends on the parity of the other sum term *)
 Lemma Zeven_plus_equiv :
   forall (n m : Z), Z.Even n -> (Z.Even m <-> Z.Even (m + n)).
 Proof.
@@ -184,7 +184,8 @@ Proof. done. Qed.
     Regarding the transition which leads to the final state, it technically could be
     included, but we choose to model this way, in order to be consistent
     with the subsequent example, where adding the last transition makes a qualitative
-    difference to the trace *)
+    difference to the trace
+*)
 Definition parity_trace1_init : list (transition_item ParityVLSM) :=
   [ Build_transition_item parity_label (Some 4) (8, 4) (Some 8)
   ; Build_transition_item parity_label (Some 2) (8, 2) (Some 4) ].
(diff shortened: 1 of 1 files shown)


theories/VLSM/Core/Examples/Parity.v
```diff
@@ -18,20 +18,26 @@ From VLSM.Core Require Import VLSM PreloadedVLSM VLSMProjections.
...
 
+(** Parity is preserved by taking opposite *)
 Lemma Zeven_unary_minus :
   forall n : Z, Z.Even n <-> Z.Even (-n).
 Proof. by intros n; split; intros [p Hp]; exists (-p); lia. Qed.
 
+(** Even subtrahend and difference implies even minuend 
+    This lemma will be useful when proving the final result of this section, because of the way 
+    we defined the transitions in the Parity VLSM *)
+(** Maybe also renamed to be more suggestive in the sense that is about a difference? *)
```
palmskog, line 29: Subtrahend and minuend I think are reasonable latin words, but they are considered archaic in English. Maybe better to use something simple like "left-hand side"?

The "Maybe also" seems like a TODO comment, we probably don't want to include that one.

theories/VLSM/Core/Examples/Parity.v
```diff
@@ -173,6 +180,9 @@ Proof. done. Qed.
 
 (** *** Example of a valid trace *)
 
+(** The initial state cannot be included to this definition, because, since there is no transition reaching this state, it cannot be expressed in the below manner
+    Regarding the transition which leads to the final state, it technically could be included, but we choose to model this way, in order to be consistent
+    with the subsequent example, where adding the last transition makes a qualitative difference to the trace *)
```
palmskog, line 185: Please make the lines shorter (around 80 characters per line).

palmskog (commented):
Looks good in general, but some nitpicks.

theories/VLSM/Core/Examples/Parity.v
palmskog: The only name improvement I can think of is to actual include the subtraction as part of the name, e.g., `Zeven_sub_preserve_parity`.

DafinaTrufas: Yes, this is what I was thinking about. So should I change it?

theories/VLSM/Core/Examples/Parity.v
```diff
@@ -40,6 +46,7 @@ Proof.
     by rewrite Zeven_equiv, <- Hparity, <- Zeven_equiv.
 Qed.
 
+(** Parity is preserved by taking difference *)
```
wkolowski, line 48:
```suggestion
(** Parity is preserved by subtraction. *)
```

theories/VLSM/Core/Examples/Parity.v
```diff
@@ -18,20 +18,26 @@ From VLSM.Core Require Import VLSM PreloadedVLSM VLSMProjections.
...
+(** Parity is preserved by taking opposite *)
 Lemma Zeven_unary_minus :
   forall n : Z, Z.Even n <-> Z.Even (-n).
 Proof. by intros n; split; intros [p Hp]; exists (-p); lia. Qed.
 
-Lemma Zeven_preserve_parity :
+(** Even right-hand side and difference implies even left-hand side 
+    This lemma will be useful when proving the final result of this section, because of the way 
+    we defined the transitions in the Parity VLSM *)
+(** Maybe also renamed to be more suggestive in the sense that is about a difference? *)
```
wkolowski: Don't forget to remove this TODO-like comment.

theories/VLSM/Core/Examples/Parity.v
```diff
@@ -51,13 +58,13 @@ Proof.
   by apply Zeven_unary_minus.
 Qed.
 
+(** Even sum term implies parity of the sum depends on the parity of the other sum term *)
```
wkolowski, line 60: This comment is not really helpful. It just paraphrases the lemma statement in natural language, which is much less clear than the formal statement itself.

theories/VLSM/Core/Examples/Parity.v
palmskog, line 29:
```suggestion
```

palmskog (approved):
@DafinaTrufas you can merge after removing the "Maybe also" comment.

theories/VLSM/Core/Examples/Parity.v
```diff
@@ -173,6 +180,12 @@ Proof. done. Qed.
 
 (** *** Example of a valid trace *)
 
+(** The initial state cannot be included to this definition, because, since there is no
+    transition reaching this state, it cannot be expressed in the below manner
+    Regarding the transition which leads to the final state, it technically could be
+    included, but we choose to model this way, in order to be consistent
+    with the subsequent example, where adding the last transition makes a qualitative
+    difference to the trace *)
```
wkolowski, line 187:
```suggestion
(**
  The initial state cannot be included to this definition, because, since there is no
  transition reaching this state, it cannot be expressed in the below manner
  Regarding the transition which leads to the final state, it technically could be
  included, but we choose to model this way, in order to be consistent
  with the subsequent example, where adding the last transition makes a qualitative
  difference to the trace.
*)
```

theories/VLSM/Core/Examples/Parity.v
```diff
@@ -380,7 +394,7 @@ Proof.
   by apply vlsm_incl_pre_loaded_with_all_messages_vlsm.
 Qed.
 
-(** *** Constrained messages are even integers *)
+(** *** Constrained messages are strictly positive even integers *)
```
wkolowski, line 396: I think "positive" suffices here.

```suggestion
(** *** Constrained messages are positive even integers *)
```

PR #274, opened 2023-07-21
Updating the build guide with instructions for coq-lsp

DafinaTrufas:


palmskog (commented):
I think it would be more helpful if we split the instructions into one part for VsCoq and one part for coq-lsp. For example:
```
## VsCoq Extension

<instructions go here>

## Coq LSP Extension

<instructions go here>
```
It would also be nice if you could specify the `code --install-extension` command for VsCoq, so the instructions are analogous.

Finally, we already have the switch creation instructions above, can you do a cross-reference to that subsection instead of repeating `opam switch ...`?

BUILDING.md
```diff
@@ -148,4 +148,18 @@ Then, install the following extensions in the Remote WSL environment:
 - [the **VsCoq** extension](https://marketplace.visualstudio.com/items?itemName=maximedenes.vscoq)
 - [the **Fast Unicode Math Characters** extension](https://marketplace.visualstudio.com/items?itemName=GuidoTapia2.unicode-math-vscode)
 
+As an alternative to **VsCoq**, you can use [the **coq-lsp** extension](https://coq.discourse.group/t/ann-coq-lsp-0-1-7/1996). This extension requires Coq 8.17.1 version, so in order to use it, you will need to create a dedicated opam switch and install the project's dependencies on it:
```
palmskog, line 151:
```suggestion
As an alternative to **VsCoq**, you can use [**coq-lsp** extension](https://github.com/ejgallego/coq-lsp). This extension requires Coq 8.17.1 version, so in order to use it, you will need to create a dedicated opam switch and install the project's dependencies on it:
```

palmskog:
To deal with the problem of Coq versions, we can just use Coq 8.17.1 everywhere.

DafinaTrufas:
> To deal with the problem of Coq versions, we can just use Coq 8.17.1 everywhere.

So that means that it's not necessary to create a new switch for Coq-LSP anymore, right? I'm a bit confused because you said that: "we already have the switch creation instructions above, can you do a cross-reference to that subsection instead of repeating opam switch ...?".

palmskog:
Sure, yes, if we specify Coq 8.17.1 everywhere, you don't need to tell people to create a *new* switch. But you can do a crossreference to the the switch creation and say that you assume this has been done. Then you give the `opam install coq-lsp` command.

DafinaTrufas:
> I think it would be more helpful if we split the instructions into one part for VsCoq and one part for coq-lsp. For example:
> 
> ```
> ### VsCoq Extension
> 
> <instructions go here>
> 
> ### Coq LSP Extension
> 
> <instructions go here>
> ```
> 
> It would also be nice if you could specify the `code --install-extension` command for VsCoq, so the instructions are analogous.
> 
> Finally, we already have the switch creation instructions above, can you do a cross-reference to that subsection instead of repeating `opam switch ...`?

I cannot find the `code --install-extension` command for VSCoq. Is it ok to provide only the links, for all the extensions?

palmskog:
If we extrapolate from the URLs and the coq-lsp command, it should be:
```shell
code --install-extension maximedenes.vscoq
```
Maybe you can try it locally?

DafinaTrufas:
> ```shell
> code --install-extension maximedenes.vscoq
> ```

Yes, I tried it now and it works. I add the commands back.

BUILDING.md
```diff
@@ -61,10 +61,10 @@ disable the sandboxing by choosing "Y".
 
 Choose "y" in order to allow opam to modify `~/.profile`.
 
-### Install a switch for opam
+### <a name="switch-creation"></a>Install a switch for opam
```
palmskog, line 64: GitHub Markdown already automatically gives you an anchor name based on the header content, in this case it's `#install-a-switch-for-opam`.

DafinaTrufas: Ah, I didn't know. I'll fix it now.

palmskog (approved):
I did some modifications and took this mergeable form.

PR #277, opened 2023-08-11
Coding conventions for pattern matching

wkolowski:


palmskog:
@wkolowski can you also please add a mention of the 2-space convention to Definitions with linebreaks? That's a different header.

wkolowski:
@palmskog Something like it was already there ("two spaces indentation before definition body"), but I reworded it to emphasize it should only be done if a definition doesn't fit on a single line.

CODING_CONVENTIONS.md
```diff
@@ -120,7 +120,7 @@ Inductive lv_event_type : Type :=
 ### Definitions
 
 - C-style name
-- two spaces indentation before definition body
+- If the definition doesn't fit into a single line, indent its body with 2 spaces starting in the next line.
```
palmskog, line 123:
```suggestion
- if a definition doesn't fit into a single line, indent its body with 2 spaces when starting on the next line.
```

palmskog (approved):
Fine by me, but one final suggestion.

PR #282, opened 2023-09-29
Longest common prefixes and suffixes

wkolowski:
For SUMO we need a function that computes longest common suffixes of states. The simplest way to implement it is to lift an analogous function that does the same for lists.

In this PR I define a function that compute longest common prefixes and use it to implement a function that computes longest common suffixes. Both come with some handy lemmas and a specification.

### DIFF
--- a/theories/VLSM/Lib/ListExtras.v
+++ b/theories/VLSM/Lib/ListExtras.v
@@ -1565,6 +1565,15 @@ Definition element_of_filter
 
 (** ** Computing longest common prefixes and suffixes *)
 
+(** *** Longest common prefix *)
+
+Section sec_longest_common_prefix.
+
+Context
+  {A : Type}
+  `{EqDecision A}
+  .
+
 (**
   [longest_common_prefix] computes the longest common prefix of two lists
   together with the suffixes that result from removing this prefix from
@@ -1572,7 +1581,7 @@ Definition element_of_filter
   if <<l1 = l ++ r1>>, <<l2 = l ++ r2>> and <<l>> is the longest list that
   satisfies these conditions.
 *)
-Fixpoint longest_common_prefix `{EqDecision A} (l1 l2 : list A) : list A * list A * list A :=
+Fixpoint longest_common_prefix (l1 l2 : list A) : list A * list A * list A :=
 match l1, l2 with
 | [], _ => ([], l1, l2)
 | _, [] => ([], l1, l2)
@@ -1585,19 +1594,19 @@ match l1, l2 with
 end.
 
 Lemma longest_common_prefix_nil_l :
-  forall {A : Type} `{EqDecision A} (l : list A),
+  forall (l : list A),
     longest_common_prefix [] l = ([], [], l).
 Proof. done. Qed.
 
 Lemma longest_common_prefix_nil_r :
-  forall {A : Type} `{EqDecision A} (l : list A),
+  forall (l : list A),
     longest_common_prefix l [] = ([], l, []).
 Proof.
   by destruct l.
 Qed.
 
 Lemma longest_common_prefix_diag :
-  forall {A : Type} `{EqDecision A} (l : list A),
+  forall (l : list A),
     longest_common_prefix l l = (l, [], []).
 Proof.
   induction l as [| h t]; cbn; [done |].
@@ -1605,15 +1614,15 @@ Proof.
 Qed.
 
 Lemma longest_common_prefix_head :
-  forall {A : Type} `{EqDecision A} (l1 l2 : list A),
+  forall (l1 l2 : list A),
     head l1 <> head l2 -> longest_common_prefix l1 l2 = ([], l1, l2).
 Proof.
-  intros A ? [] [] Hneq; cbn in *; [done.. |].
+  intros [] [] Hneq; cbn in *; [done.. |].
   by rewrite decide_False; [| congruence].
 Qed.
 
 Lemma longest_common_prefix_app :
-  forall {A : Type} `{EqDecision A} (l l1 l2 p r1 r2 : list A),
+  forall (l l1 l2 p r1 r2 : list A),
     longest_common_prefix l1 l2 = (p, r1, r2) ->
       longest_common_prefix (l ++ l1) (l ++ l2) = (l ++ p, r1, r2).
 Proof.
@@ -1622,8 +1631,8 @@ Proof.
   by apply IHt in Heq as ->.
 Qed.
 
-Lemma longest_common_prefix_app' :
-  forall {A : Type} `{EqDecision A} (l l1 l2 : list A),
+Lemma longest_common_prefix_app_let :
+  forall (l l1 l2 : list A),
     longest_common_prefix (l ++ l1) (l ++ l2) =
       let '(p, r1, r2) := longest_common_prefix l1 l2 in (l ++ p, r1, r2).
 Proof.
@@ -1633,7 +1642,7 @@ Proof.
 Qed.
 
 Lemma longest_common_prefix_app_inv :
-  forall {A : Type} `{EqDecision A} (l1 l2 p r1 r2 : list A),
+  forall (l1 l2 p r1 r2 : list A),
     longest_common_prefix l1 l2 = (p, r1, r2) ->
       l1 = p ++ r1 /\ l2 = p ++ r2.
 Proof.
@@ -1645,8 +1654,8 @@ Proof.
   by apply IHt1 in Heq as [-> ->].
 Qed.
 
-Lemma longest_common_prefix_longest :
-  forall {A : Type} `{EqDecision A} (l1 l2 p r1 r2 : list A),
+Lemma longest_common_prefix_head_inv :
+  forall (l1 l2 p r1 r2 : list A),
     longest_common_prefix l1 l2 = (p, r1, r2) ->
       r1 = [] /\ r2 = [] \/ head r1 <> head r2.
 Proof.
@@ -1662,7 +1671,7 @@ Proof.
 Qed.
 
 Lemma longest_common_prefix_spec :
-  forall {A : Type} `{EqDecision A} (l1 l2 p r1 r2 : list A),
+  forall (l1 l2 p r1 r2 : list A),
     longest_common_prefix l1 l2 = (p, r1, r2)
       <->
     l1 = p ++ r1 /\ l2 = p ++ r2 /\ (r1 = [] /\ r2 = [] \/ head r1 <> head r2).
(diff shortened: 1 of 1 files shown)


theories/VLSM/Lib/ListExtras.v
```diff
@@ -1562,3 +1562,213 @@ Definition element_of_filter
...
+  forall {A : Type} `{EqDecision A} (l l1 l2 p r1 r2 : list A),
+    longest_common_prefix l1 l2 = (p, r1, r2) ->
+      longest_common_prefix (l ++ l1) (l ++ l2) = (l ++ p, r1, r2).
+Proof.
+  induction l as [| h t]; cbn; intros * Heq; [done |].
+  rewrite decide_True by done.
+  by apply IHt in Heq as ->.
+Qed.
+
+Lemma longest_common_prefix_app' :
```
palmskog, line 1625: I think it's bad practice to just add a prime/tick to an existing name. Unless there are established conventions, there is no way to know exactly what the prime/tick is supposed to mean. I think there are two options here:

- keep the tick/prime but add a `#[local]` attribute so the lemma can't be easily used outside the module
- come up with a better and and more descriptive name

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1562,3 +1562,213 @@ Definition element_of_filter
...
+Lemma longest_common_suffix_diag :
+  forall {A : Type} `{EqDecision A} (l : list A),
+    longest_common_suffix l l = (l, [], []).
+Proof.
+  intros.
+  unfold longest_common_suffix.
+  by rewrite longest_common_prefix_diag, reverse_involutive; cbn.
+Qed.
+
+Lemma longest_common_suffix_app' :
```
palmskog, line 1731: As for `longest_common_prefix_app'`, I think this is a bad name, and there are two me two options:

- keep the tick/prime but add a `#[local]` attribute so the lemma can't be easily used outside the module
- come up with a better and and more descriptive name

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1562,3 +1562,213 @@ Definition element_of_filter
...
+Proof.
+  induction l1 as [| h1 t1]; destruct l2 as [| h2 t2]; cbn;
+    only 1-3: intros * [= <- <- <-]; cbn; [done.. |].
+  intros p r1 r2.
+  destruct (longest_common_prefix t1 t2) as [[]] eqn: Heq.
+  destruct (decide (h1 = h2)); subst; intros [= <- <- <-]; cbn.
+  - by apply IHt1 in Heq.
+  - by rewrite decide_False.
+Qed.
+
```
palmskog, line 1717: There is already a notion of "prefix" in stdpp: https://gitlab.mpi-sws.org/iris/stdpp/-/blob/6e560df6564ceefda86ddfae12368c1ba3711ef5/stdpp/list.v#L291

To validate, please connect your definitions/results to the stdpp one.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1562,3 +1562,213 @@ Definition element_of_filter
...
+    destruct (longest_common_prefix (reverse l1) (reverse l2)) as [[]] eqn: Heq.
+    apply longest_common_prefix_spec in Heq as (Heq1 & Heq2 & H).
+    apply (f_equal reverse) in Heq1, Heq2.
+    rewrite reverse_involutive in Heq1, Heq2.
+    rewrite Heq1, Heq2, !reverse_app.
+    intros [= <- <- <-].
+    split_and!; [done.. |].
+    rewrite !last_reverse.
+    by destruct H as [[-> ->] |]; cbn; [left | right].
+Qed.
```
palmskog, line 1774: There is already a notion of "suffix" in stdpp: https://gitlab.mpi-sws.org/iris/stdpp/-/blob/6e560df6564ceefda86ddfae12368c1ba3711ef5/stdpp/list.v#L290

To validate, please connect your definitions/results to the stdpp one.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1562,3 +1562,213 @@ Definition element_of_filter
...
+(** ** Computing longest common prefixes and suffixes *)
+
+(**
+  [longest_common_prefix] computes the longest common prefix of two lists
+  together with the suffixes that result from removing this prefix from
+  these lists, i.e. <<longest_common_prefix l1 l2 = (l, r1, r2)>> if and only
+  if <<l1 = l ++ r1>>, <<l2 = l ++ r2>> and <<l>> is the longest list that
+  satisfies these conditions.
+*)
+Fixpoint longest_common_prefix `{EqDecision A} (l1 l2 : list A) : list A * list A * list A :=
```
palmskog, line 1575: Since seemingly every definition/result has an `EqDecision` instance, why not organize them inside a single section with an `EqDecision A` context? This would also make it easier to split up this file later, since it's getting very big.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1562,3 +1562,213 @@ Definition element_of_filter
...
+  induction l1 as [| h1 t1]; destruct l2 as [| h2 t2]; cbn;
+    only 1-3: intros * [= <- <- <-]; cbn; [done.. |].
+  intros p r1 r2.
+  destruct (longest_common_prefix t1 t2) as [[]] eqn: Heq.
+  destruct (decide (h1 = h2)); subst; intros [= <- <- <-]; cbn.
+  - by apply IHt1 in Heq.
+  - by rewrite decide_False.
+Qed.
+
+Definition longest_common_suffix `{EqDecision A} (l1 l2 : list A) : list A * list A * list A :=
```
palmskog, line 1718: As for the prefix definition, why not put all suffix definitions inside a single section for easier maintenance and less repetition?

palmskog (commented):
Please address my comments.

theories/VLSM/Lib/ListExtras.v
wkolowski: The lemma is certainly not local. As for a better name, any ideas? `longest_common_prefix_app` is quite a good name, but there are two versions of it: one has an equational hypothesis and one that uses a `let` instead. For now I'll rename the latter into something like `longest_common_prefix_app_let`, but it's not very satisfying. @traiansf suggested `longest_common_prefix_app_rew`, but it's not perfect either.

Do you know if there is any naming convention used anywhere for lemmas that are derived from others by replacing some part of it with equational hypotheses?

theories/VLSM/Lib/ListExtras.v
wkolowski: I'm renaming it into `longest_common_prefix_app_let` for now.

theories/VLSM/Lib/ListExtras.v
wkolowski: Ok, I will put it in a separate section with `A` and `EqDecision A` in a `Context` block.

theories/VLSM/Lib/ListExtras.v
palmskog: `longest_common_prefix_app_let` is fine by me, at least it can't easily be mistaken for `longest_common_prefix_app`. Works also for suffix.

theories/VLSM/Lib/ListExtras.v
wkolowski: Same as above, it'll get its own section.

theories/VLSM/Lib/ListExtras.v
wkolowski: Sadly it can't be used for computation, as it lives in `Prop`.

I added 3 lemmas to relate `longest_common_prefix` with stdpp's `prefix` and another 3 to do the same for `longest_common_suffix` and stdpp's `suffix`.

theories/VLSM/Lib/ListExtras.v
wkolowski: Sadly it can't be used for computation, as it lives in `Prop`.

I added 3 lemmas to relate `longest_common_prefix` with stdpp's `prefix`.

theories/VLSM/Lib/ListExtras.v
wkolowski: As above, I added 3 lemmas to relate `longest_common_suffix` and stdpp's `suffix`.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1715,33 +1724,75 @@ Proof.
...
+Proof.
+  intros * Hprefix.
+  apply longest_common_prefix_spec in Hprefix as (-> & -> & _).
+  by split; apply prefix_app_r.
+Qed.
+
+Lemma longest_common_prefix_is_longest :
+  forall (l l1 l2 : list A),
+    prefix l l1 -> prefix l l2 ->
+      let '(p, r1, r2) := longest_common_prefix l1 l2 in prefix l p.
```
palmskog, line 1740: Viewed in isolation, this is fine as lemma statement, but the name is confusing to me - I don't see anything that is "longest". If you don't intend to state that other prefixes are as long or longer in terms of list length, then maybe `prefix_longest_common_prefix` would be better.

theories/VLSM/Lib/ListExtras.v
```diff
@@ -1772,3 +1823,36 @@ Proof.
...
+Proof.
+  intros * Hsuffix.
+  apply longest_common_suffix_spec in Hsuffix as (-> & -> & _).
+  by split; apply suffix_app_r.
+Qed.
+
+Lemma longest_common_suffix_is_longest :
+  forall (l l1 l2 : list A),
+    suffix l l1 -> suffix l l2 ->
+      let '(p, r1, r2) := longest_common_suffix l1 l2 in suffix l p.
```
palmskog, line 1840: As per above, `suffix_ longest_common_suffix` might be a better name.

palmskog (approved):
This looks much more reasonable now, but take a look at some final comments.

theories/VLSM/Lib/ListExtras.v
wkolowski: Well, it's saying that all prefixes of `l1` and `l2` are also prefixes of their common prefix `p`, so `p` must be the longest such prefix. Maybe my perception is biased by category theory, but I think it's not that confusing.

Anyway, I'll take your name, since this lemma probably won't be used anywhere.

theories/VLSM/Lib/ListExtras.v
wkolowski: Changed.

PR #286, opened 2023-10-03
Replace "ram" with "constrained" in names

wkolowski:
I noticed that both `ram_state_prop` and `constrained_state_prop` are used to refer to the notion of valid state in the preloaded vlsm, and since "constrained" is the more up to date term that we prefer, I did some refactoring to get rid of the old "ram" naming.


traiansf (approved):
Looks OK to me.

PR #287, opened 2023-10-06
Define `adr2idx`

wkolowski:
BaseELMO already had a function `ELMO_A : Address -> index` which could be used to find an index of the component with the given address. The problem with it was that it was defined with a default value instead of returning an option, and this needed to be changed to prove an additional lemma which was needed for SUMO.


traiansf (approved):
Looks good to me

PR #289, opened 2023-10-16
Parity-primes

traiansf:
Infinite composition for parity using prime numbers and a validation result for components.

PR #290, opened 2023-10-23
Finitely supported functions

traiansf:
Extracting from PR #289 the part about finitely supported functions.

### DIFF
--- a/theories/VLSM/Lib/FinSuppFn.v
+++ b/theories/VLSM/Lib/FinSuppFn.v
@@ -6,33 +6,35 @@ From VLSM.Lib Require Import Preamble StdppExtras ListExtras.
 (** * Finitely supported functions *)
 
 (**
+  A function is finitely supported if its [support] is [Finite].
+*)
+
+(**
   The support of a function is the type of elements of its domain that are not
   mapped to the specified inhabitant of its codomain.
 
   Note that for [nat] and [Z] the default [inhabitant] is [0].
 
-  Note that we require the codomain to have decidable equality, because
-  we want to make our lives easier with regard to proving equality of elements
+  Note that we require the codomain to have decidable equality, because it
+  allows for a straight-forward approach to proving equality of elements
   of the support.
-
-  Then a function is finitely supported if its support is [Finite].
 *)
-Definition supp `(f : A -> B) `{Inhabited B} `{EqDecision B} : Type :=
+Definition support `(f : A -> B) `{Inhabited B} `{EqDecision B} : Type :=
   dsig (fun a => f a <> inhabitant).
 
 Section sec_fin_supp_fn.
 
 Context
-  {A : Type}
+  {A B : Type}
   `{EqDecision A}
-  `{EqDecision B}
   `{Inhabited B}
+  `{EqDecision B}
   .
 
-Definition fin_supp (f : A -> B) `{!Finite (supp f)} : list A :=
-  map proj1_sig (enum (supp f)).
+Definition fin_supp (f : A -> B) `{!Finite (support f)} : list A :=
+  map proj1_sig (enum (support f)).
 
-Lemma elem_of_fin_supp (f : A -> B) `{!Finite (supp f)} :
+Lemma elem_of_fin_supp (f : A -> B) `{!Finite (support f)} :
   forall (a : A), a ∈ fin_supp f <-> f a <> inhabitant.
 Proof.
   unfold fin_supp.
@@ -43,20 +45,17 @@ Proof.
     by exists (dexist a Ha); split; [| apply elem_of_enum].
 Qed.
 
-Lemma not_elem_of_fin_supp (f : A -> B) `{!Finite (supp f)} :
+Lemma not_elem_of_fin_supp (f : A -> B) `{!Finite (support f)} :
   forall (a : A), a ∉ fin_supp f <-> f a = inhabitant.
 Proof.
-  pose (Helem := elem_of_fin_supp f).
-  firstorder.
-  apply dec_stable.
-  contradict H0.
-  by apply Helem.
+  intro a; rewrite elem_of_fin_supp.
+  by destruct (decide (f a = inhabitant)); itauto.
 Qed.
 
-Lemma fin_supp_NoDup `(f : A -> B) `{!Finite (supp f)} : NoDup (fin_supp f).
+Lemma fin_supp_NoDup `(f : A -> B) `{!Finite (support f)} : NoDup (fin_supp f).
 Proof. by apply dsig_NoDup, NoDup_enum. Qed.
 
-Lemma fin_supp_proper (f g : A -> B) `{!Finite (supp f)} `{!Finite (supp g)} :
+Lemma fin_supp_proper (f g : A -> B) `{!Finite (support f)} `{!Finite (support g)} :
   f = g -> fin_supp f ≡ₚ fin_supp g.
 Proof.
   intros Heq.
@@ -64,13 +63,14 @@ Proof.
   by intro; rewrite !elem_of_fin_supp; subst.
 Qed.
 
-#[export] Instance fin_supp_fn_eq_dec (f g : A -> B) `{!Finite (supp f)} `{!Finite (supp g)} :
+#[export] Instance fin_supp_fn_eq_dec
+  (f g : A -> B) `{!Finite (support f)} `{!Finite (support g)} :
   Decision (f = g).
 Proof.
   destruct (@finset_equiv_dec A (listset A) _ _ _ _ _ _ _ _ _
     (list_to_set (fin_supp f)) (list_to_set (fin_supp g))) as [Heqv | Hneqv]; cycle 1.
-  - right; intros ->; apply Hneqv.
-    by intros a; rewrite !elem_of_list_to_set, !elem_of_fin_supp.
+  - right; contradict Hneqv.
+    by rewrite fin_supp_proper.
   - destruct (decide (set_Forall (fun a => f a = g a) (list_to_set (C := listset A) (fin_supp f))))
       as [Hall | Hall]; [| by right; contradict Hall; subst].
     left; extensionality a.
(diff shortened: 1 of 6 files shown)


wkolowski (commented):
I will finish my PR with suggestions tomorrow. In the meantime, I opened a PR #291 with an alternative definition of finitely supported function, which you might want to look at and eventually steal :)

theories/VLSM/Lib/StdppExtras.v
```diff
@@ -633,3 +633,50 @@ Proof.
   - by apply Hsub, elem_of_app; left.
   - by apply Hsub, elem_of_app; right.
 Qed.
+
+#[export] Instance finset_equiv_dec `{FinSet A C} : RelDecision (≡@{C}).
```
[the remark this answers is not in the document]
traiansf: couldn't find it.

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,540 @@
...
+  to that function.
+*)
+Record fin_dom_fn (A B Supp : Type) := mk_fin_dom_fn
+{
+  fin_dom_f : A -> B ;
+  fin_dom : Supp ;
+}.
+Arguments fin_dom {A B Supp}%type_scope  f : assert.
+Arguments fin_dom_f {A B Supp}%type_scope f _ : assert.
+Arguments mk_fin_dom_fn {A B Supp}%type_scope fin_dom_f%function_scope fin_dom : assert.
```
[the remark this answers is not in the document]
traiansf: Not sure.  I'm just an intelligent-monkey in this case; I saw that we're usually using `assert` with `Arguments`,  so I did the same here.

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,540 @@
...
+
+
+(** * Finitely supported functions *)
+
+(**
+  We model finitely supported functions as consisting of a total function
+  and a finite set of elements from the function's source type (called the domain),
+  satisfying that an element is in the domain iff its image through the
+  function is different than the [inhabitant].
+  Note that for [nat] and [Z] the [inhabitant] is <<0>>.
```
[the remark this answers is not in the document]
traiansf: I think we agreed in the meeting that it's ok to use `Inhabitant`

traiansf:
Based on the discussion during the meeting today I've completely rewrirten finitely supported functions.
It is quite similar to the one proposed by @wkolowski  in PR #291 except that I'm using a dsig to ensure equality reduces to element equality.

theories/VLSM/Lib/FinSuppFn.v
wkolowski, line 1: Are you sure you want `Defined` in the places where you use it?

Also, there is some relying on autogenerated names, you should fix it.

theories/VLSM/Lib/StdppExtras.v
```diff
@@ -441,38 +441,6 @@ Proof.
...
-  (l : list A)
-  (Hs : Forall P l)
-  (xP : dsig P)
-  : xP ∈ list_annotate P l Hs <-> (` xP) ∈ l.
-Proof.
-  split; [by apply elem_of_list_annotate_forget |].
-  destruct xP as [x Hpx]; cbn.
-  by induction 1; cbn; rewrite elem_of_cons, dsig_eq; cbn; auto.
-Qed.
-
```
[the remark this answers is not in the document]
traiansf: well, I've added a couple of lemmas on list annotate, and not knowing that `elem_of_list_annotate` exists I actually rewrote it again; hence, I decided to move all results concerning `list_annotate` together

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,456 @@
...
+*)
+Section sec_fin_supp_fn.
+
+Context
+  `{EqDecision A}
+  `{EqDecision B}
+  `{Inhabited B}
+  .
+
+Definition fin_supp `(f : A -> B) `{!Finite (supp f)} : list A :=
```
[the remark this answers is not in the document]
traiansf: That might be so; however, those hypotheses would still have to be in the context, because we want two finitely supported functions with the same domain and codomain to share the same instances of `Inhabited` and `EqDecision`.

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,454 @@
...
+  The support of a function is the type of elements of its domain that are not
+  mapped to the specified inhabitant of its codomain.
+
+  Note that for [nat] and [Z] the default [inhabitant] is [0].
+
+  Note that we require the codomain to have decidable equality, because it
+  allows for a straight-forward approach to proving equality of elements
+  of the support.
+*)
+Definition support `(f : A -> B) `{Inhabited B} `{EqDecision B} : Type :=
```
palmskog, line 22:
```suggestion
Definition support `(f : A -> B) `{Inhabited B, EqDecision B} : Type :=
```
We can compress a bit here and elsewhere.

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,454 @@
...
+  forall (a : A), a ∉ fin_supp f <-> f a = inhabitant.
+Proof.
+  intro a; rewrite elem_of_fin_supp.
+  by destruct (decide (f a = inhabitant)); itauto.
+Qed.
+
+Lemma fin_supp_NoDup `(f : A -> B) `{!Finite (support f)} : NoDup (fin_supp f).
+Proof. by apply dsig_NoDup, NoDup_enum. Qed.
+
+Lemma fin_supp_proper (f g : A -> B) `{!Finite (support f)} `{!Finite (support g)} :
```
palmskog, line 58:
```suggestion
Lemma fin_supp_proper (f g : A -> B) `{!Finite (support f), !Finite (support g)} :
```
A bit more succinct.

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,454 @@
...
+  - destruct (decide (set_Forall (fun a => f a = g a) (list_to_set (C := listset A) (fin_supp f))))
+      as [Hall | Hall]; [| by right; contradict Hall; subst].
+    left; extensionality a.
+    destruct (decide (a ∈ fin_supp f)) as [| Hf]; [by apply Hall, elem_of_list_to_set |].
+    destruct (decide (a ∈ fin_supp g)) as [| Hg]; [by apply Hall; rewrite Heqv, elem_of_list_to_set |].
+    apply not_elem_of_fin_supp in Hf, Hg.
+    by congruence.
+Qed.
+
+Definition empty_supp_fn : A -> B :=
```
palmskog, line 83: I'm not a big fan of having the `fn` prefix/suffix everywhere. Why not call it just `empty_supp` and change other names accordingly?

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,454 @@
...
+    (elements (fin_supp_nat_fn_add_supp f1 f2)).
+Proof.
+  unfold fin_supp_nat_fn_add_supp; apply Forall_forall; intros a.
+  rewrite elem_of_elements, elem_of_union, !elem_of_list_to_set,
+    !elem_of_fin_supp.
+  by unfold fin_supp_nat_fn_add; cbn; lia.
+Qed.
+
+#[export] Program Instance fin_supp_nat_fn_add_has_finn_supp
+  (f1 f2 : A -> nat) `{!Finite (support f1)} `{!Finite (support f2)}
```
palmskog, line 316:
```suggestion
  (f1 f2 : A -> nat) `{!Finite (support f1), !Finite (support f2)}
```

theories/VLSM/Lib/FinSuppFn.v
```diff
@@ -0,0 +1,454 @@
...
+Proof.
+  by intros; rewrite fin_supp_nat_fn_add_comm, fin_supp_nat_fn_add_succ_l, fin_supp_nat_fn_add_comm.
+Qed.
+
+(**
+  To be able to prove things by induction on finitely supported functions on
+  naturals we define the following inductive property and then we show that it
+  holds for all such functions.
+*)
+Inductive FinSuppNatFn : (A -> nat) -> Type :=
```
palmskog, line 388: If we really need to use something like the `Fn` suffix here, why not just `Fun`?

theories/VLSM/Lib/Preamble.v
```diff
@@ -725,3 +725,18 @@ Program Definition not_lt_plus_dec {m n} (Hnlt : ~ n < m) : {k | k + m = n} :=
   exist _ (n - m) _.
 Next Obligation.
 Proof. by cbn; lia. Qed.
+
+Definition update_fn `{EqDecision A} `(f : A -> B) (a : A) (b : B) : A -> B :=
```
palmskog, line 729: Why can't we just call it `update`? See [here](https://github.com/uwplse/StructTact/blob/master/theories/Update.v) in StructTact, where the same function is called just `update`.

theories/VLSM/Lib/StdppExtras.v
```diff
@@ -633,3 +601,63 @@ Proof.
...
+  - intros Hsum i Hi.
+    apply sum_list_with_in with (f := f) in Hi.
+    by lia.
+  - induction l; intros Hall; cbn; [done |].
+    rewrite Hall by left.
+    rewrite IHl; [done |].
+    by intros; apply Hall; right.
+Qed.
+
+Lemma dsig_NoDup `(P : A -> Prop) `{Pdec : forall a, Decision (P a)} :
```
palmskog, line 650:
```suggestion
Lemma dsig_NoDup_map `(P : A -> Prop) `{Pdec : forall a, Decision (P a)} :
```
We can at least hint at that this has to do with lists.

palmskog (approved):
I think this is by far the cleanest proposed formalization of finitely supported functions so far, and avoids introducing more typeclasses nicely. I only have some minor comments.

theories/VLSM/Lib/FinSuppFn.v
wkolowski: `empty_supp` suggests "empty support", but that's not the same thing as a function with empty support.

wkolowski:
@palmskog It's indeed better than the previous versions, but I'm afraid the story is not over. There's another version of it in PR #293 and now we need to choose which is better. I added you as a reviewer there, as you might want to take a look. I myself will try to compare both approaches later today.

PR #297, opened 2023-10-30
Rename lemmas about `UMO_reachable`

wkolowski:
Originally in UMO there were some lemmas about `valid_state_prop`, but later @bmmoore generalized them to use `UMO_reachable` (which he defined). However, the names of the lemmas were not changed, and so they were super confusing. I decided to change them to better reflect what the lemma is about.

Changes:
- from `valid_state_prop_inv_Ri` to `UMO_reachable_inv`
- from `valid_state_prop_addObservation_Ri` to `UMO_reachable_addObservation_inv`
- from `valid_state_prop_addObservation_Ri_Send` to `UMO_reachable_addObservation_inv_Send_state`
- from `valid_state_prop_addObservation_Ri_Send'` to `UMO_reachable_addObservation_inv_Send`
- from `addObservation_message` to `UMO_reachable_addObservation_inv_message`
- from `valid_state_prop_addObservations_Ri` to `UMO_reachable_addObservations_inv`

### DIFF
--- a/theories/VLSM/Core/ELMO/UMO.v
+++ b/theories/VLSM/Core/ELMO/UMO.v
@@ -936,7 +936,7 @@ Proof.
   by apply addObservation_inj in Hadd as [_ ->].
 Qed.
 
-Lemma UMO_reachable_addObservation_inv_Send P :
+Lemma UMO_reachable_addObservation_inv_Send_state P :
   forall (s : State) (m : Message),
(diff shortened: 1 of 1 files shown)


traiansf (approved):

bmmoore (approved):
Good change. I agree I should have renamed this from the beginning, for the "valid_state_prop" substrings that reflected the propositions involved in the statement.

PR #298, opened 2023-10-30
Swap definitions of `constrained_state_prop` and `constrained_state_prop_alt`

wkolowski:
The definition of `constrained_state_prop` was analogous to that of `valid_state_prop` (going through an inductive definition, etc.), whereas `constrained_state_prop_alt` was just `valid_state_prop` applied to the preloaded VLSM. This was unfortunate, because the alternative definition was used much more often than the "nonalternative" one.

I swapped these two definitions, so now `constrained_state_prop` is `valid_state_prop` applied to preloaded VLSM, whereas `constrained_state_prop_alt` is the one that directly mimics the definition of `valid_state_prop`. I also moved the definitions of the constrained things from `VLSM.v` to `PreloadedVLSM.v`.

### DIFF
--- a/theories/VLSM/Core/Composition.v
+++ b/theories/VLSM/Core/Composition.v
@@ -682,9 +682,6 @@ Lemma constrained_state_prop_component
     constrained_state_prop (free_composite_vlsm IM) cs ->
     constrained_state_prop (IM i) (cs i).
 Proof.
-  intros Hcsp.
-  rewrite constrained_state_prop_alt_equiv in Hcsp |- *.
-  unfold constrained_state_prop_alt in Hcsp |- *.
   by eapply valid_state_project_preloaded_to_preloaded_free.
 Qed.
 
--- a/theories/VLSM/Core/ELMO/ELMO.v
+++ b/theories/VLSM/Core/ELMO/ELMO.v
@@ -134,7 +134,7 @@ Lemma full_node_VLSM_reachable
     forall (l : Label) (s : State) (m : Message),
       valid V l (s, Some m) -> full_node s m) :
   forall (s : State),
-    constrained_state_prop_alt V s ->
+    constrained_state_prop V s ->
     UMO_reachable full_node s.
 Proof.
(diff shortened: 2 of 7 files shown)


palmskog (commented):
I don't think I'm the right person to review/approve this PR, since it basically concerns high-level issues with VLSMs rather than Coq-specific things.

Wouldn't it be better to come up with unique names for the two concepts rather than just adding the `_alt` prefix to one of them?

traiansf:
If we want to have more distinct names, I would let the most used one be constrained_states and name the alternate something like constrained_state_direct_definition, because it's based on the direct definition given in the paper

palmskog:
This is going into bikeshed territory, but then how about `constrained_states` and `constrained_state_direct`?

wkolowski:
Maybe let's just merge this swap PR and leave changing names for later? Coming up with names is the hard part here...

palmskog:
@wkolowski but you are already proposing a renaming in this PR. If there is no consensus on what these things should be called (and by extension, what they mean), then this PR amounts to renaming `bar` into `foo` and `foo` into `bar`, and I don't see what is gained by doing this. In a "proper" renaming PR, there would be some documentation explaining the differences and use cases of these definitions.

wkolowski:
@palmskog It's not really a renaming, just a "swap". The intention is to take one of the two definitions, the one that is used much more often and for which proofs are a bit shorter, and assign it the name `constrained_state_prop`, which is shorter than the one with the `_alt` suffix and also does not look as suspicious.

I think for this particular use case (having an alternative definition which is very close to what's in the paper), the `_alt` suffix is not that bad. Maybe you could replace it with `_direct` (short for "direct definition"), but I'm not sure if that would help a lot.

wkolowski:
@palmskog To be clear, in the next PR from this series I will use all this `constrained` all over UMO/MO/ELMO. I think it would be really silly if these hundreds of occurrences were all `_alt` instead of the ones without the suffix.

PR #300, opened 2023-10-31
Change naming patterns in UMO/MO/ELMO

wkolowski:
In `PreloadedVLSM.v` I added more definitions of `constrained` things that correspond to the `valid` ones we already have.

In UMO, I changed lemmas from using `valid` concepts and `Ri` (the preloaded UMO component) to using the `constrained` analogues and renamed them accordingly. For lemmas about the non-preloaded component `Ui`, I removed the `Ui` suffix from their names. I also changed some comments that relate to all this.

Changes for MO are similar, except that here the preloaded component was called `RMi`.

For ELMO, I changed the lemma statements from using `Ri` (preloaded ELMO component) to using `constrained` stuff, but often kept the names since they already had "constrained" somewhere in them. I also removed `Ri` from some `Context`s as they were not useful, and fixed a few proofs that required an additional use of the `red` tactic.

### DIFF
--- a/theories/VLSM/Core/ELMO/UMO.v
+++ b/theories/VLSM/Core/ELMO/UMO.v
@@ -305,10 +305,6 @@ Qed.
 (** ** Every valid state contains a unique valid trace leading to it
 
   To prove this, we will need some basic properties of UMO components.
-
-  For the rest of this section, we will work with two UMO components,
-  one (named [Ui]) for dealing with valid states, and
-  another (named [Ri]) for dealing with reachable states.
 *)
 
 Section sec_UMOComponent_lemmas.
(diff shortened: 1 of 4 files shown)


theories/VLSM/Core/ELMO/UMO.v
```diff
@@ -305,19 +305,16 @@ Qed.
...
 *)
 
 Section sec_UMOComponent_lemmas.
 
 (**
   [Ui] is a notation for an [UMOComponent] of address [i].
 
-  The "R" in [Ri] stands for "reachability", as it will be used to state and
-  prove lemmas and theorems which talk about reachability.
+  [Ri] is a notation for an [UMOComponent] of address [i] preloaded with all
```
traiansf, line 315: DO you still need this comment?

traiansf (approved):

PR #301, opened 2023-10-31
Remove the last remnants of the _Ri/_RMi naming

wkolowski:
The last lemmas to use this naming were ones that established an inclusion from the component to the preloaded component. I dropped them and used `vlsm_incl_pre_loaded_with_all_messages_vlsm` directly.

ELMO was using the preloaded component `Ri` in a few places where the non-preloaded one would suffice. I changed that. Also, I reindented some lines in ELMO that were too long after the recent changes.


traiansf (approved):
Looks OK to me, but it seems you also have to convince Coq 8.16

PR #302, opened 2023-11-01
Parity primes reloaded

traiansf:
- Changed constraint to require that messages are larger than 1 to match the new version in the paper
- Extracted finitely supported function as separate module
- Extracted properties about primes and products of powers into `NatExtras` module
- Added an additional example from the paper requiring that messages are even.


theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -453,31 +458,29 @@ Qed.
...
   forall (m : ParityMessage),
-    (exists (j : Z), m = multiplier * j) -> m > 0 ->
+    (exists (j : Z), m = multiplier * j) -> m > multiplier ->
     constrained_message_prop ParityVLSM m.
 Proof.
   intros Hgt0 m (j & Hj) Hmgt0.
   unfold constrained_message_prop_alt, can_emit.
   exists (j, Some j), parity_label, 0.
-  repeat split; [| | by lia.. | by cbn; do 2 f_equal; lia]; cycle 1.
+  repeat split; cycle 1.
```
wkolowski, line 467: Why the `cycle`?

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -741,160 +625,124 @@ Proof.
...
+      * apply initial_state_is_valid.
+        intros j; cbn; unfold ParityComponent_initial_state_prop.
+        by case_decide; lia.
+      * done.
+      * by rewrite decide_True; [lia |].
+      * by lia.
+      * cbn; f_equal; extensionality j.
+        rewrite decide_True by done.
+        destruct (decide (n = j)); subst; state_update_simpl; [by lia |].
+        by rewrite decide_False.
```
wkolowski, line 730: Maybe this should be a lemma?

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -1077,3 +925,235 @@ Proof.
...
+(**
+  For any prime number, its corresponding component in the [primes_vlsm_composition]
+  is a validator for that composition, i.e., all of its constrained transitions
+  can be lifted to valid transitions for the composition.
+*)
+Theorem component_projection_validator_prop_primes :
+  forall (p : primes),
+    component_projection_validator_prop
+      (indexed_parity_vlsms (fun p : primes => ` p))
+      (const (const True)) p.
```
wkolowski, line 972: I find it much clearer to just write

```suggestion
      (fun _ _ => True) p.
```

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -1077,3 +925,235 @@ Proof.
...
+(** *** A (slightly) constrained composition of primes *)
+
+(**
+  To show how a composition constraint affects validation, we here add a very
+  simple constraint, that the received message must be even, to the composition.
+*)
+Inductive EvenConstraint :
+  label primes_vlsm_composition ->
+  state primes_vlsm_composition * option Z -> Prop :=
+| even_constraint : forall l s m, Z.Even m -> EvenConstraint l (s, Some m).
```
wkolowski, line 996: Why an `Inductive` instead of a `Definition`? And why is it so badly formatted?

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -1077,3 +925,235 @@ Proof.
...
+      by case_decide; lia.
+    + by apply initial_message_is_valid.
+    + by rewrite decide_True; [lia |].
+    + by lia.
+    + by exists 1.
+    + cbn; f_equal; [| by f_equal; lia].
+      extensionality j.
+      rewrite decide_True by done.
+      destruct (decide (dexist n Hp = j)); subst; state_update_simpl; [by lia |].
+      by rewrite decide_False.
```
wkolowski, line 1031: Maybe a separate lemma?

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -1077,3 +925,235 @@ Proof.
...
+      by case_decide; lia.
+    + by apply Hind.
+    + by rewrite decide_True; [lia |].
+    + by lia.
+    + by exists q.
+    + cbn; f_equal; [| by f_equal; lia].
+      extensionality j.
+      rewrite decide_True by done.
+      destruct (decide (dexist m Hp = j)); subst; state_update_simpl; [by lia |].
+      by rewrite decide_False.
```
wkolowski, line 1049: Maybe a separate lemma?

wkolowski (approved):
Looks ok technically. I assume you know what you are doing with all these logic changes at the beginning. I opened a PR with suggestions #303

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -280,28 +283,28 @@ Proof.
...
+Definition parity_trace2_init_last_state : ParityState := multiplier + 1.
 
 Definition parity_trace2_last_state : ParityState :=
   destination parity_trace2_last_item.
 
 (** The given trace is valid without the last transition. *)
 
 Proposition parity_valid_transition_1' :
-  multiplier > 0 ->
+  multiplier > 1 ->
```
palmskog, line 305: The assumption `multiplier > 1` seems to be appear in a lot of places. Maybe it should be a section `Hypothesis`?

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -1077,3 +922,228 @@ Proof.
 Qed.
 
 End sec_parity23.
+
+(** ** A composition for all primes
```
palmskog, line 926: Since this is partially aimed at new users, maybe it would be better with
> A VLSM composition for all primes

palmskog (approved):
This looks very good now, but I have a few minor comments, and we should preserve compatibility with Stdpp 1.8.0 if it's not too much trouble.

theories/VLSM/Core/Examples/ParityComposition.v
palmskog: If we find ourselves using `fun _ _ => X` a lot, how about we just add a definition `const2` for this?

theories/VLSM/Core/Examples/ParityComposition.v
palmskog: The good thing about defining `Inductive` is that they don't unfold when using `simpl`, `cbn`, etc. So if we want to prevent accidental unfolding of a predicate, it can be a good idea. But if a regular `Definition` suffices, it can require less work to juggle when proving things.

traiansf: Well, my main reason here was to avoid an extra match and a `False` case

palmskog: @traiansf but you shouldn't need a `False` case, right? It should just be something like:
```coq
Definition EvenConstraint l s m : Prop := Z.Even m -> EvenConstraint l (s, Some m).
```
Or am I missing something?

traiansf: Would something like this even work without `Inductive` ?
but anyway, I need it to have that particular signature (using label and state*option message) to pass it to the composition operation.

palmskog: I mean, if you only have a single constructor of the `Inductive`, and it has no recursive argument, then it's going to be equivalent (in the sense of `<->` and type signatures) as the same predicate using `Definition`.

wkolowski: I think using pattern matching this would be

```coq
match om with
| None => False
| Some m' => Z.Even m'
end.
```

So yeah, maybe `Inductive` is not that bad of a choice here.

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -659,13 +644,34 @@ Section sec_free_composition.
...
   free_composite_vlsm (indexed_parity_vlsms multipliers).
 
+#[local] Lemma free_parity_composite_vlsm_emits_multiplier :
+  forall (m : ParityMessage),
+    valid_message_prop free_parity_composite_vlsm m ->
+    m >= 2 ->
+  forall (n : index),
+  input_valid_transition free_parity_composite_vlsm (existT n parity_label)
+  (λ j : index, if decide (n = j) then m + 1 else 1, Some m)
+  (λ _ : index, 1, Some (multipliers n * m)).
```
wkolowski, line 660:
```suggestion
(fun j : index => if decide (n = j) then m + 1 else 1, Some m)
(fun _ : index => 1, Some (multipliers n * m)).
```

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -742,6 +733,18 @@ Proof.
   - by intros [? []]; eapply composition_valid_messages_powers_of_mults_left.
 Qed.
 
+Lemma composition_valid_message_ge_2
+  (Hmpos : forall (i : index), multipliers i > 1) (m : ParityMessage) :
+  valid_message_prop free_parity_composite_vlsm m ->
+  m >= 2.
```
wkolowski, line 739:
```suggestion
  valid_message_prop free_parity_composite_vlsm m -> m >= 2.
```

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -742,6 +733,18 @@ Proof.
...
+Proof.
+  intro Hv.
+  apply composition_valid_messages_powers_of_mults in Hv
+    as (i & Hsupp & ->); [| done].
+  cut (fsfun_prod multipliers i > 1); [by lia |].
+  by apply prod_powers_gt; [lia| ..].
```
wkolowski, line 745: Can't you do it without the `cut`?

```suggestion
  by apply prod_powers_gt; [lia | ..].
```

theories/VLSM/Core/Examples/ParityComposition.v
```diff
@@ -993,6 +996,30 @@ Inductive EvenConstraint :
...
 
+#[local] Lemma even_constrained_primes_composition_emits_multiplier :
+  forall (m : ParityMessage),
+    valid_message_prop even_constrained_primes_composition m ->
+    m >= 2 ->
+    Z.Even m ->
+  forall (n : primes),
+  input_valid_transition even_constrained_primes_composition (existT n parity_label)
+  (λ j : primes, if decide (n = j) then m + 1 else 1, Some m)
+  (λ _ : primes, 1, Some (` n * m)).
```
wkolowski, line 1007: `fun` instead of `λ`

theories/VLSM/Lib/NatExtras.v
```diff
@@ -3,6 +3,10 @@ From stdpp Require Import prelude finite.
 From VLSM.Lib Require Import EquationsExtras.
 From VLSM.Lib Require Import Preamble FinSuppFn StdppExtras ListExtras.
 
+(** TODO: remove this once we remove support for stdpp 1.8.0 *)
+#[export] Instance Z_mul_assoc : Assoc (=) (Z.mul).
+Proof.  by intros x y z; apply Z.mul_assoc. Qed.
```
wkolowski, line 8: Do I have OCD or normal people can't distinguish at a glance between 1 and 2 spaces?
```suggestion
Proof. by intros x y z; apply Z.mul_assoc. Qed.
```

theories/VLSM/Lib/NatExtras.v
```diff
@@ -159,3 +164,225 @@ Proof.
...
+
+(** ** Products of (natural) powers of integers *)
+
+Section sec_prod_powers.
+
+Context
+  `{EqDecision index}
+  `(multipliers : index -> Z)
+  .
+
```
palmskog, line 176: Hmm, to get rid of all the `%Z` in this section, maybe we should use `Open Scope Z_scope` in a future PR.

PR #307, opened 2023-11-03
Clean up syntax in VLSM.v

wkolowski:
`VLSM.v` had a couple of layers of different indentation/code layout conventions accrued over time, which looked rather silly. I standardized it to look mostly like the code we're writing now. I tried not to touch proofs, but I probably did a few times... When reviewing this PR, it might be better to look at the actual file rather than the diff.


traiansf (approved):
I'll aprove this one, but maybe wait a bit longer before you do another one in the same vein.

PR #308, opened 2023-11-06
Renaming and a direct definition for validity

traiansf:


theories/VLSM/Core/PreloadedVLSM.v
```diff
@@ -612,18 +612,26 @@ Definition constrained_state_message_prop :=
 
 End sec_constrained_defs.
 
-(** ** Alternate definitions to constrained traces, states and messages
+(** ** Alternate (direct) definitions for constrained traces, states and messages
```
wkolowski, line 615:
```suggestion
(** ** Direct definitions of constrained traces, states and messages
```

theories/VLSM/Core/PreloadedVLSM.v
```diff
@@ -612,18 +612,26 @@ Definition constrained_state_message_prop :=
...
 
-  "Constrained" concepts can also be defined in an alternative way, by mimicking
-  definitions for the corresponding "valid" concepts.
+  In the above formalization, "valid" concepts are defined first, then
+  "constrained" ones are derived from them, expressed as validity within the
+  [pre_loaded_with_all_messages_vlsm].
+
+  In this section we retrieve the original definitions and we show them
```
wkolowski, line 621:
```suggestion
  In this section we retrieve the original definitions and we show them
```

This sounds rather bad without any context. Where do the original definitions come from?

theories/VLSM/Core/Equivocation.v
```diff
@@ -2853,12 +2853,14 @@ Lemma all_pre_traces_to_valid_state_are_valid_free
 Proof.
   apply pre_traces_with_valid_inputs_are_valid in Htr; [done |].
   apply valid_trace_last_pstate in Htr as Hspre.
-  intros.
+  apply Forall_forall; intros.
+  destruct (input x) as [m |] eqn: Hm; [| by apply option_valid_message_None].
```
wkolowski, line 2857: Where do these changes come from?

traiansf: I made a slight change to the statement of `pre_traces_with_valid_inputs_are_valid`.

theories/VLSM/Core/PreloadedVLSM.v
```diff
@@ -693,4 +698,135 @@ Proof.
...
+  them from the "constrained" notions, and we showing them equivalent with the
+  ones defined in the VLSM module.
+*)
+
+(**
+  A valid trace is a constrained trace whose [input]s are all valid messages;
+  a valid message is either an initial message or an [output] of a valid trace.
+*)
+Inductive
+  finite_valid_trace_init_to_def :
```
palmskog, line 714: The `_def` suffix is confusing here, what does it signify? I think any developer or user will struggle to understand the difference between `x` and `x_def`. Can we use something else more descriptive, such as `_abstract`, `_abs`?

theories/VLSM/Core/PreloadedVLSM.v
```diff
@@ -693,4 +698,135 @@ Proof.
     by eexists _, _, _, _.
 Qed.
 
-End sec_constrained_defs_alt.
+(** ** Indirect definitions of valid traces, messages, and states
+
```
palmskog, line 702: Maybe I'm missing something, but I thought we already had indirect definitions elsewhere in the codebase? Doesn't this mean there are now three different definitions (direct, indirect, and other)?

theories/VLSM/Core/PreloadedVLSM.v
```diff
@@ -693,4 +698,135 @@ Proof.
...
+Lemma valid_state_prop_def_right_impl :
+  forall (s : state X), valid_state_prop X s -> valid_state_prop_def s.
+Proof.
+  intros s Hs.
+  apply valid_state_has_trace in Hs as (is & tr & Htr).
+  by exists is, tr; apply finite_valid_trace_init_to_def_equiv.
+Qed.
+
+Lemma valid_state_prop_def_equiv :
+  forall (s : state X), valid_state_prop_def s <-> valid_state_prop X s.
```
palmskog, line 825: This is particularly puzzling as a statement of a lemma viewed in isolation. One would have to look inside both `valid_state_prop_def` and `valid_state_prop` to figure out the differences.

palmskog (commented):
I really like the addition of `_direct` and documentation of this, but the `_def` ones I have a much harder time with, see my comments.

theories/VLSM/Core/PreloadedVLSM.v
traiansf: for valid traces, we only have the recursive definition based on valid transition, which in turn uses the recursive definition where for valid states and message pairs are defined together.
We do have an alternate mutually recursive definition for valid states and messages but we didn't have any of the definitions which are currently in the paper.

theories/VLSM/Core/PreloadedVLSM.v
wkolowski: It's no different with `_alt` or any other suffix.

palmskog: I completely disagree with this. `_alt` at least gives the idea that `x_alt` is an alternative to `x` (not necessarily recommended). If all suffixes are "the same", when would any kind of naming conventions ever be meaningful?

traiansf:
Since there was no contention about the renamings concerning `constrained` stuff, I decided to move the other stuff to a separate PR.

palmskog (approved):
With all the "controversial" changes moved to a different PR, this is a very obvious improvement, thanks.

PR #309, opened 2023-11-07
Some additional comments for Lib modules

wkolowski:
Sometime in summer @palmskog suggested that maybe we should have higher comment coverage. I did some work on it and then forgot about it. Since we will have a release soon, it's a good time to revive this work.

In this PR, I added comments to some definitions (i.e. non-opaque things) in `VLSM.Lib`. I also wrote some sections headers, changed some existing comments and did a few more documentation-related things.

### DIFF
--- a/theories/VLSM/Lib/Itauto.v
+++ b/theories/VLSM/Lib/Itauto.v
@@ -1,6 +1,22 @@
 From Cdcl Require Export Itauto.
 
-(** * Constructive Itauto tactic *)
+(** * Constructive Itauto tactic
+
+  This module contains a workaround that prevents the itauto tactic from using
+  classical logic.
+
+  The problem is that itauto uses classical logic by default if the current
+  module requires (even if only transitively) any module that uses classical
+  logic. For example, most modules about real numbers from the standard library
+  cause itauto to use excluded middle basically each time it's called.
+
+  The solution is to redefine the itauto tactic to avoid doing this.
+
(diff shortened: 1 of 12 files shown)


traiansf (approved):
fine with me.

theories/VLSM/Lib/Itauto.v
```diff
@@ -1,6 +1,22 @@
...
+  module requires (even if only transitively) any module that uses classical
+  logic. For example, most modules about real numbers from the standard library
+  cause itauto to use excluded middle basically each time it's called.
+
+  The solution is to redefine the itauto tactic to avoid doing this.
+
+  Every time we want to use itauto, we should import it from the current module
+  instead of directly from the Cdcl library.
+
+  TODO: this workaround should be removed after the problem is fixed upstream.
```
palmskog, line 18: This problem was actually fixed upstream for the Itauto version for Coq 8.18. When the minimum version of Coq we require is 8.18, all we need to do is put `#[export] Set Itauto Classic.` in `Ctauto.v`, replacing the other content (and have `Itauto.v` be a pure export). Please be specific about the Coq version in this comment.

palmskog (approved):
This is fine, but I think documentation of utility functions and results is far down in the priority list from VLSM-specific documentation.

theories/VLSM/Lib/Itauto.v
wkolowski: Changed.

PR #310, opened 2023-11-07
Remove `list_prefix` and `list_suffix`

wkolowski:
I noticed that `list_prefix` and `list_suffix` were (almost) identical to `firstn` and `skipn` from stdpp/stdlib, so I replaced the former with the latter. We also had some lemmas proved which were already provided by stdpp/stdlib, so I removed these.

### DIFF
--- a/theories/VLSM/Lib/StdppExtras.v
+++ b/theories/VLSM/Lib/StdppExtras.v
@@ -244,15 +244,16 @@ Proof.
   by etransitivity; [| apply lookup_lt_is_Some].
 Qed.
 
-Lemma list_suffix_lookup
+Lemma skipn_lookup
   {A : Type}
   (s : list A)
   (n : nat)
   (i : nat)
   (Hi : n <= i)
-  : list_suffix s n !! (i - n) = s !! i.
+  : skipn n s !! (i - n) = s !! i.
 Proof.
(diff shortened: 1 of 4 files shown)


theories/VLSM/Lib/StdppExtras.v
```diff
@@ -244,15 +244,16 @@ Proof.
   by etransitivity; [| apply lookup_lt_is_Some].
 Qed.
 
-Lemma list_suffix_lookup
+Lemma skipn_lookup
```
traiansf, line 247: Mâine this could go into ListExtras to be next to the other skipn lemmas

traiansf (approved):

theories/VLSM/Lib/StdppExtras.v
palmskog: I agree, `skipn` is from Stdlib. Ideally `StdppExtras.v` is reserved for definitions/results specifically about stdpp.

palmskog (approved):
Looks good to me, but please address the comments by myself and Traian.

theories/VLSM/Lib/StdppExtras.v
wkolowski: Ok, I will move lemmas about `firstn` and `skipn` into `ListExtras.v`.

PR #312, opened 2023-11-08
Differentiate parity examples

palmskog:
This is one of steps toward a better tutorial experience for VLSM users. We distinguish the `ParitySingle` and `ParityComposition` modules by name and description, and use different names inside each module for parity-related VLSM definitions.

Maybe someone can suggest a better place for the `app_valid_tran` tactic. In any case, it shouldn't be duplicated with minor variations.

### DIFF
--- a/theories/VLSM/Core/Examples/Parity.v
+++ b/theories/VLSM/Core/Examples/Parity.v
@@ -1,10 +1,10 @@
 From stdpp Require Import prelude finite.
 From VLSM.Core Require Import VLSM PreloadedVLSM VLSMProjections.
 
-(** * Parity Single VLSM
+(** * Parity VLSM
 
   This module demonstrates some basic notions of the VLSM framework. We define
-  a (single) parity VLSM that stores a tuple and continually decrement one of the
+  a (single) parity VLSM that stores a tuple and continually decrements one of the
   tuple's elements while a constraint is checked at each step. The name originates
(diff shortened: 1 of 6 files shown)


traiansf (approved):

wkolowski (approved):
Since you're doing renamings anyway, maybe you want to rename `RadixComponent` into `radix_component` to make it consistent with the recent renamings for ELMO? Also, maybe it should just be called `radix`? `radix_component` makes little sense, since the composition is not called `radix`.

palmskog:
I did some renaming in the examples to avoid having `Component` in a bunch of names (it probably didn't help). This makes the naming conventions quite close to ELMO.

theories/VLSM/Core/Examples/Parity.v
```diff
@@ -110,18 +97,18 @@ Definition ParityType : VLSMType ParityMessage :=
...
+Definition Parity_transition
   (l : ParityLabel) (s : ParityState) (om : option ParityMessage)
   : ParityState * option ParityMessage :=
   match om with
   | Some j  => ((s.1, s.2 - j), Some (2 * j))
   | None    => (s, None)
   end.
 
-Definition ParityComponentValid (l : ParityLabel) (st : ParityState)
+Definition ParityValid (l : ParityLabel) (st : ParityState)
```
wkolowski, line 111: Maybe rename this to `Parity_valid`?

PR #314, opened 2023-11-08
Remove `firstn` and `skipn`

wkolowski:
After the recent changes, we are now using `firstn` and `skipn` directly instead of `list_prefix` and `list_suffix`. Since `firstn` and `skipn` are displayed as `take` and `drop` in proof mode anyway, I replaced all uses of the former with the latter.


palmskog (commented):
`take` and `drop` are notations introduced in stdpp. I'm fine with embracing them, but then I think the `take`/`drop` lemmas in `ListExtras.v` should move to `StdppExtras.v`. (I was unaware of the plans for this followup in my review of #310.)

palmskog (approved):
Based on the discussion elsewhere, I'm approving this, and we can address the "utility module organization problem" in the future.

PR #315, opened 2023-11-09
Refactor `longest_common_prefix` and `longest_common_suffix`

wkolowski:
stdpp has `max_prefix` and `max_suffix`, which are the same as `longest_common_prefix` and `longest_common_suffix` which were previously implemented manually, so I removed the latter. Some proofs were also simplified, but most lemmas couldn't be removed as they are missing from stdpp.


palmskog (approved):
Fine by me.

PR #317, opened 2023-11-10
Replace `map_option` with `omap` from stdpp

wkolowski:
`ListExtras.v` had a definition of `map_option`, which was definitionally equal to `omap` from stdpp. I replaced all uses of the former with the latter and removed some duplicated lemmas. I also simplified some proofs for `cap_option` which was defined using `map_option`.


traiansf (approved):

palmskog (approved):
Fine by me.

PR #318, opened 2023-11-13
Refactor ListExtras

wkolowski:
`ListExtras.v` was quite messy. I changed the formatting to be consistent with what we prefer now. I simplified most proofs, some significantly so. I removed quite a few lemmas that were already provided by stdlib or stdpp.


traiansf (approved):
reasonable

PR #320, opened 2023-11-13
add formalization of Paxos using VLSMs under Examples

palmskog:
Merging this PR will mean that we only support Coq-std++ 1.9.0 or later.

Unfortunately, there is no Coq Platform release with this version of Coq-std++ yet. We can add back the Platform-related building documentation when the Platform release appears.


BUILDING.md
wkolowski, line 1: Are changes to this file intended? And if yes, why put them in this PR, which should be about moving code between repos?

palmskog: I can only quote my PR description:

> Merging this PR will mean that we only support Coq-std++ 1.9.0 or later.

> Unfortunately, there is no Coq Platform release with this version of Coq-std++ yet. We can add back the Platform-related building documentation [in `BUILDING.md`] when the Platform release appears.

palmskog: > But still, are the build changes related to Paxos in any way?

The Paxos code added here is not compatible with Coq-std++ 1.8.0. This is what triggers the changes to non-Coq files.

traiansf:
Would it be hard to alter the Paxos definition to maintain compatibility with std++ 1.8.0?
Would it make sense to have a compatibility module with missing functions from std++ 1.9.0?

palmskog:
@traiansf the changes in stdpp 1.9.0 in the definitions used by Paxos were such that I consider it unfeasible to be compatible with both stdpp 1.8.0 and 1.9.0. It's not just a question of missing definitions, but also fundamental changes in how typeclasses are defined and in typeclass resolution.

Since stdpp 1.9.0 is the version used in the next Coq Platform that will be released soon for Coq 8.18, it doesn't make much sense to me to only support stdpp 1.8.0. Recall that stdpp 1.9.0 supports older Coq (back to Coq 8.16).

wkolowski (approved):
Since Paxos was pretty heavily reviewed before and didn't change in the meantime, it's probably ok. The only suggestions I have (in PR #324) is some record formatting.

theories/Examples/Paxos/Consensus.v
```diff
@@ -0,0 +1,94 @@
...
+  .
+
+Definition consensus_state : Type := VSet.
+Definition consensus_message : Type := False.
+
+Inductive consensus_action : Type :=
+| Decided (v : V).
+
+(** None for a no-op transition. *)
+Definition consensus_label : Type := option consensus_action.
```
palmskog, line 39: @bmmoore usually, an `Inductive` with only a single constructor might be better written as a record. But then the label is an `option`. Would the following make sense:
```coq
Inductive consensus_label : Type := Decided (v : V) |Undecided.
```
... or was there some specific reason not to do it this way? We don't need to change it right now, but maybe good to know the background.

bmmoore: Allowing stuttering transitions independent of any details of the definition of a particular distributed system is essential to the strategy of proof by refinement, and seems to be a common feature of TLA+ models in general. Having the `option` as a separately layer is supposed to represent that stuttering is a possibility independent of the details of the actual transitions.

I define the transitions as an `Inductive` to reflect that it's a sum of the alternative sorts of steps in the transition relation, even if in this particular system it is a degenerate sum of a single alternative.

PR #321, opened 2023-11-14
Added some results and documentation for Multiply and PrimesComposition.

traiansf:


theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+Proof. by exists two; unshelve eexists (exist _ 2 _); cbn; lia. Qed.
+
+Lemma composition_23_3_initial : composite_initial_message_prop components_23 3.
+Proof. by exists three; unshelve eexists (exist _ 3 _); cbn; lia. Qed.
+
+Lemma free_composition_23_valid_messages_powers_of_23_right :
+  forall (m : multiplying_message),
+    valid_message_prop free_composition_23 m ->
+      exists p2 p3 : nat, (p2 + p3 >= 1)%nat /\
+        m = 2 ^ Z.of_nat p2 * 3 ^ (Z.of_nat p3).
```
palmskog, line 635: Hmm, can't we let `p2` and `p3` be of type `Z` here? All the `Z.of_nat` stuff is quite distracting. Can't one have something like: `exists p2 p3 : Z, p2 >= 0 /\ p3 >= 0 /\ p2 + p3 >= 1 /\ ....`?

theories/Examples/Tutorial/Multiply.v
```diff
@@ -767,34 +1082,92 @@ Proof.
...
+Qed.
+
+(**
+  From any composite state whose components are even natural numbers there is
+  a valid trace reaching [state00].
+*)
+Theorem parity_composite_23_reachable_00 :
+  forall (s2 s3 : nat),
+    in_futures parity_composition_23
+      (state_23 (2 * Z.of_nat s2) (2 * Z.of_nat s3)) state00.
```
palmskog, line 1146: How about `s2 s3 : Z`?

traiansf:
We could; the main reason for switching to nat was to omit extra hypotheses and  to make induction more straight-forward. I'll change them to Z if you think it's better.

palmskog:
@traiansf if this was somewhere else, I might agree with you, but here (in the most basic tutorial) I'd think we want to avoid drawing attention to the not-so-obvious differences between `nat` and `Z` - if it can be avoided. The induction shouldn't be that much harder, right?

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+  revert Hvsm Hom; generalize (Some m) as om; intros.
+  clear m; induction Hvsm using valid_state_message_prop_ind.
+  - destruct Hom as [m ->]; cbn in *; subst.
+    destruct Hom0 as ([] & [] & <-); cbn in *.
+    + by exists 1%nat, 0%nat; split; lia.
+    + by exists 0%nat, 1%nat; split; lia.
+  - destruct l as [[] []];
+      (destruct om as [m |]; [| done]);
+      (unshelve edestruct IHHvsm2 as (p2 & p3 & Hgt1 & Heq); [done |]);
+      cbn in *; inversion Ht; subst; cbn in *.
```
wkolowski, line 649: This looks rather complicated. Isn't there a simpler way?

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+    + exists (S p2), p3.
+      split; [by lia |].
+      rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
```
wkolowski, line 652:
```suggestion
      by rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
```

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+    + exists p2, (S p3).
+      split; [by lia |].
+      rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
```
wkolowski, line 655:
```suggestion
      by rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
```

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+      repeat split.
+      * by apply initial_state_is_valid; intros []; cbn; red; lia.
+      * replace (3 ^ Z.of_nat (S p3)) with (2 ^ Z.of_nat 0 * 3 ^ Z.of_nat (S p3))
+          by lia.
+        by eapply Hind; cycle 1; [done | lia..].
+      * by rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
+      * by cbn; lia.
+      * cbn; f_equal;
+          [| by f_equal; rewrite !Nat2Z.inj_succ, !Z.pow_succ_r; lia].
+        by extensionality i; destruct i; state_update_simpl; cbn; lia.
```
wkolowski, line 680: I think using another level of bullets would look better here.

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+      repeat split.
+      * by apply initial_state_is_valid; intros []; cbn; red; lia.
+      * replace (2 ^ Z.of_nat (S p2)) with (2 ^ Z.of_nat (S p2) * 3 ^ Z.of_nat 0)
+          by lia.
+        by eapply Hind; cycle 1; [done | lia..].
+      * by rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
+      * by cbn; lia.
+      * cbn; f_equal;
+          [| by f_equal; rewrite !Nat2Z.inj_succ, !Z.pow_succ_r; lia].
+        by extensionality i; destruct i; state_update_simpl; cbn; lia.
```
wkolowski, line 695: Again.

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+        Some (2 ^ Z.of_nat p2 * 3 ^ Z.of_nat (S p3))),
+      (existT two multiply_label), state01.
+    repeat split.
+    * by apply initial_state_is_valid; intros []; cbn; red; lia.
+    * by eapply Hind; cycle 1; [done | lia..].
+    * by rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
+    * by cbn; lia.
+    * cbn; f_equal;
+        [| by f_equal; rewrite !Nat2Z.inj_succ, !Z.pow_succ_r; lia].
+      by extensionality i; destruct i; state_update_simpl; cbn; lia.
```
wkolowski, line 708: Again.

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+  - by apply free_composition_23_valid_messages_powers_of_23_right.
+  - intros (? & ? & ? & ->).
+    by apply free_composition_23_valid_messages_powers_of_23_left.
+Qed.
+
+Definition state_update_23 (s : composite_state components_23)
+  (i : index23) (z : Z) (j : index23) : state (components_23 j) :=
+  if (decide (i = j)) then
+    match j with two => z | three  => z end
+  else s j.
```
wkolowski, line 731:
```suggestion
  if decide (i = j)
  then
    match j with
    | two => z
    | three  => z
    end
  else s j.
```

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
 Definition free_composition_23 : VLSM multiplying_message :=
   free_composite_vlsm components_23.
 
+Lemma composition_23_2_initial : composite_initial_message_prop components_23 2.
+Proof. by exists two; unshelve eexists (exist _ 2 _); cbn; lia. Qed.
```
wkolowski, line 626: I think it would be more readable for newcomers when written this way:
```suggestion
Lemma composition_23_2_initial :
  composite_initial_message_prop components_23 2.
Proof.
  by exists two; unshelve eexists (exist _ 2 _); cbn; lia.
Qed.
```
I think you should apply this style throughout.

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+  if (decide (i = j)) then
+    match j with two => z | three  => z end
+  else s j.
+
+Lemma state_update_23_eq (s : composite_state components_23) (i : index23) (z : Z) :
+  zproj (state_update_23 s i z) i = z.
+Proof. by destruct i; cbn; unfold state_update_23; rewrite decide_True. Qed.
+
+Lemma state_update_23_neq (s : composite_state components_23) (i : index23) (z : Z) :
+  forall (j : index23), j <> i -> state_update_23 s i z j = s j.
```
wkolowski, line 738: I don't like this mixing of parameters and `forall`, since you are using `intros` anyway. And you could just put the `j` along with the `i`.

Maybe some of the parameters should go into a `Context`?

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+
+Lemma state_update_23_twice (s : composite_state components_23) (i : index23) :
+  forall (z z' : Z),
+    state_update_23 (state_update_23 s i z) i z' = state_update_23 s i z'.
+Proof.
+  intros; extensionality j; unfold state_update_23; destruct i, j; cbn.
+  - by rewrite !decide_True.
+  - by rewrite !decide_False.
+  - by rewrite !decide_False.
+  - by rewrite !decide_True.
```
wkolowski, line 749: Silly.
```suggestion
  by intros; extensionality j; unfold state_update_23; case_decide; subst.
```

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+Qed.
+
+Definition label_23 (i : index23) : composite_label components_23.
+Proof.
+  exists i.
+  by destruct i; exact ().
+Defined.
+
+Definition multiplier_23 (i : index23) : Z :=
+  match i with two => 2 | three => 3 end.
```
wkolowski, line 775:
```suggestion
Definition multiplier_23 (i : index23) : Z :=
  match i with
  | two => 2
  | three => 3
  end.
```

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+
+Lemma free_composition_23_reachable_0 :
+  forall (s : composite_state components_23), valid_state_prop free_composition_23 s ->
+  forall (i : index23), 2 <= zproj s i ->
+    in_futures free_composition_23 s (state_update_23 s i 0).
+Proof.
+  intros s Hs i Hi.
+  remember (zproj s i) as si; revert s Hs Heqsi.
+    pose (P (si : Z) :=
```
wkolowski, line 784:
```suggestion
  pose (P (si : Z) :=
```

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,6 +622,235 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+    forall (s : composite_state components_23),
+      valid_state_prop free_composition_23 s ->
+      si = zproj s i ->
+    in_futures free_composition_23 s (state_update_23 s i 0)).
+  cut (P si); [done |].
+  revert si Hi; apply Zlt_lower_bound_ind; subst P; cbn.
+  intros si Hind Hlt s Hs Heqsi.
+  destruct (decide (4 <= si)); [| destruct (decide (si = 3))].
+  - pose (s' := state_update_23 s i (si - 2)).
+    cut (input_valid_transition free_composition_23 (label_23 i) (s, Some 2) (s', Some (2 * multiplier_23 i))).
```
wkolowski, line 794:
```suggestion
    cut (input_valid_transition free_composition_23 (label_23 i)
      (s, Some 2) (s', Some (2 * multiplier_23 i))).
```

theories/Examples/Tutorial/Multiply.v
```diff
@@ -634,8 +863,103 @@ Definition parity_constraint_23
...
+      split; [by lia |].
+      rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
+Qed.
+
+Lemma parity_composition_23_valid_messages_powers_of_23_left :
+  forall (p2 p3 : nat), (p2 >= 1)%nat ->
+    valid_message_prop parity_composition_23 (2 ^ Z.of_nat p2 * 3 ^ (Z.of_nat p3)).
+Proof.
+  induction p2; [by inversion 1 |].
+  intros p3 _; destruct p2; [clear IHp2; induction p3 |].
```
wkolowski, line 913: This is very fishy. You should use induction in the next bullet and then indent again.

theories/Examples/Tutorial/Multiply.v
```diff
@@ -634,8 +863,103 @@ Definition parity_constraint_23
...
+    repeat split.
+    + by apply initial_state_is_valid; intros []; cbn; red; lia.
+    + by apply IHp2; lia.
+    + by rewrite Nat2Z.inj_succ, Z.pow_succ_r; lia.
+    + by cbn; lia.
+    + apply parity_constraint_23_even; exists (2 ^ Z.of_nat p2 * 3 ^ Z.of_nat p3).
+      by rewrite !Nat2Z.inj_succ, !Z.pow_succ_r; lia.
+    + cbn; f_equal;
+        [| by f_equal; rewrite !Nat2Z.inj_succ, !Z.pow_succ_r; lia].
+      by extensionality i; destruct i; state_update_simpl; cbn; lia.
```
wkolowski, line 939: Again.

theories/Examples/Tutorial/Multiply.v
```diff
@@ -634,8 +863,103 @@ Definition parity_constraint_23
...
+    valid_message_prop parity_composition_23 m <->
+      m = 3 \/
+      exists p2 p3 : nat, (p2 >= 1)%nat /\
+        m = 2 ^ Z.of_nat p2 * 3 ^ (Z.of_nat p3).
+Proof.
+  split.
+  - by apply parity_composition_23_valid_messages_powers_of_23_right.
+  - intros [-> | (? & ? & ? & ->)];
+      [| by apply parity_composition_23_valid_messages_powers_of_23_left].
+    by apply initial_message_is_valid, composition_23_3_initial.
```
wkolowski, line 958:
```suggestion
  - intros [-> | (? & ? & ? & ->)].
    + by apply initial_message_is_valid, composition_23_3_initial.
    + by apply parity_composition_23_valid_messages_powers_of_23_left.
```

wkolowski (approved):
Looks ok. I opened a PR with suggestions: #323

theories/Examples/Tutorial/Multiply.v
traiansf: There might be, but I couldn't find it...

theories/Examples/Tutorial/Multiply.v
```diff
@@ -622,28 +622,398 @@ Definition zproj (s : composite_state components_23) (i : index23) : Z :=
...
+  (z : Z)
+  .
+
+Definition state_update_23 (j : index23) : state (components_23 j) :=
+  if (decide (i = j)) then
+    match j with
+    | two => z
+    | three  => z
+    end
+  else s j.
```
palmskog, line 757:
```suggestion
  if decide (i = j) then z else s j.
```
The `match` seems superfluous when the returned value is the same?

theories/Examples/Tutorial/Multiply.v
```diff
@@ -767,34 +1133,106 @@ Proof.
...
+    + by apply initial_message_is_valid, composition_23_2_initial.
+    + by destruct i; cbn in Hsi |- *; lia.
+    + by apply parity_constraint_23_even; exists 1; lia.
+    + destruct i; cbn; f_equal; symmetry.
+      * by apply state_update_23_two.
+      * by apply state_update_23_three.
+Qed.
+
+(**
+  From any composite state whose components are even natural numbers there is
```
palmskog, line 1204:
```suggestion
  From any composite state whose components are even non-negative integers there is
```

palmskog (approved):
Looks good to me now, just two small final suggestions.

theories/Examples/Tutorial/Multiply.v
traiansf: yes, so it would seem; however, Coq doesn't seem to be able to infer that the types for each index are the same until the index is destructed...
It seems an unfortunate side effect of working with indexed families of VLSMs which we usually don't notice because each in the family is basically the same VLSM (with slightly different parameters). However, in this case, since we decided not to parameterize the VLSMs we have to work around this issue.
Any suggestions are welcome.

palmskog: I think something called a "cast" can be used for this. But it's more advanced stuff that may confuse people. So fine with me to leave this here as-is.

wkolowski: @palmskog Can you elaborate?

wkolowski (approved):
After "manual merging" it's still ok.

PR #326, opened 2023-11-16
Improve readability in multiply and primes tutorials

palmskog:


theories/Examples/Tutorial/PrimesComposition.v
```diff
@@ -509,11 +511,11 @@ Qed.
 Lemma even_constrained_primes_composition_valid_message_char (m : Z) :
   valid_message_prop even_constrained_primes_composition m
     <->
-  (m > 1)%Z /\ Z.Even m \/ prime m.
+  ((m > 1 /\ Z.Even m) \/ prime m).
```
traiansf, line 514: you can also drop the parentheses here

traiansf (approved):

theories/Examples/Tutorial/PrimesComposition.v
palmskog: @traiansf but don't these parentheses help people read the statement properly? In my view, it's often hard to parse any use of `\/` without parentheses. But if you insist, I can change to drop parentheses.

wkolowski: I think the parentheses should stay.

wkolowski (approved):

theories/Examples/Tutorial/PrimesComposition.v
traiansf: Sorry, I misread. they are fine.

PR #330, opened 2023-11-17
Refactor muddy children

wkolowski:
This PR is mostly about making the proofs conform with our coding conventions.

One thing that remains to be done is splitting the very long proofs into lemmas, which I might try soon, but probably won't do if it takes too long.

### DIFF
--- a/theories/Examples/Tutorial/MuddyChildrenRounds.v
+++ b/theories/Examples/Tutorial/MuddyChildrenRounds.v
@@ -297,13 +297,11 @@ Lemma MC_no_equivocation_inductive_equiv :
   forall s m, MC_no_equivocation s m <-> MC_no_equivocation_inductive s m.
 Proof.
   split; destruct m; cbn.
-  - destruct (s msg_index0) eqn: Heq; destruct st_rs0; cbn; [|done].
-    destruct r; intros [[Hs Hr] | [Hs Hr]].
-    + by subst; eapply MC_no_equivocation_inductive_msg_eq; eauto.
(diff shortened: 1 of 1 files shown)


traiansf (approved):
fine with me. Splitting the long proofs would also be a significant improvement, given that this is a tutorial example.

theories/Examples/Tutorial/MuddyChildrenRounds.v
```diff
@@ -730,19 +731,20 @@ Lemma non_initial_state_has_init_tr (is s : composite_state MCVLSM)
...
+  setoid_rewrite elem_of_list_singleton.
+  destruct (decide (MC_initial_state_prop (s i)));
+    [| by destruct (IHHtr n) as (? & ? & ?); eauto].
+  destruct Ht as [(_ & _ & Hv & _) Ht], l as [j lj]; cbn in Ht, Hv.
+  destruct (MC_transition j lj (s j) iom) as [si' om'] eqn: Htj.
+  inversion Ht; subst; clear Ht.
+  destruct (decide (i = j)); subst; state_update_simpl; [| done].
+  unfold MC_initial_state_prop in m.
+  destruct (s j); cbn in *; subst; inversion Hv; subst.
+  by eauto.
```
palmskog, line 747: I don't think this is an improvement. The changed script is just as long but relies much more on setoid rewriting, which heavily depends on typeclasses in context and is not expected to be as stable over time as regular rewriting. I think you should roll this back.

PR #332, opened 2023-11-22
Turn some asserts into standalone lemmas in Muddy Children example

wkolowski:
This PR only deals with the `assert`s  for which I was pretty sure everything will go well. I will try do deal with the remaining ones in a separate PR.

### DIFF
--- a/theories/Examples/Tutorial/MuddyChildrenRounds.v
+++ b/theories/Examples/Tutorial/MuddyChildrenRounds.v
@@ -785,7 +785,7 @@ Proof.
 Qed.
 
 Lemma MC_component_invariant_helper_from_constraint :
-  forall {s : composite_state MCVLSM} {i : index} {r : nat} (status : ChildStatus),
+  forall (s : composite_state MCVLSM) (i : index) (r : nat) (status : ChildStatus),
     MC_composite_invariant s ->
     MC_constraint (existT i receive)
       (s, Some {| msg_index := i; msg_round := r; msg_status := status |}) ->
--- a/theories/Examples/Tutorial/MuddyChildrenRounds.v
+++ b/theories/Examples/Tutorial/MuddyChildrenRounds.v
@@ -776,7 +776,8 @@ Qed.
 
 Lemma consistent_valid_state_prop :
   forall (s : composite_state MCVLSM) (i : index),
-    valid_state_prop MC_composite_vlsm s -> st_rs (s i) <> None ->
+    valid_state_prop MC_composite_vlsm s ->
+    st_rs (s i) <> None ->
       consistent s.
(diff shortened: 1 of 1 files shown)


theories/Examples/Tutorial/MuddyChildrenRounds.v
```diff
@@ -749,6 +749,60 @@ Proof.
...
+
+Lemma MC_component_invariant_helper_from_constraint :
+  forall (s : composite_state MCVLSM) (i : index) (r : nat) (status : ChildStatus),
+    MC_composite_invariant s ->
+    MC_constraint (existT i receive)
+      (s, Some {| msg_index := i; msg_round := r; msg_status := status |}) ->
+    MC_component_invariant_helper
+    {|
+      st_obs := st_obs (s i); st_rs := Some {| rs_round := r; rs_status := status |}
+    |} (MuddyUnion s).
```
palmskog, line 795: Almost everywhere else in the file, we use the record constructors directly, such as `mkRS x y`. This was originally done to make pattern matching in Equations more obvious. I think we should follow the convention with constructors here. It will also make the code shorter.

theories/Examples/Tutorial/MuddyChildrenRounds.v
```diff
@@ -749,6 +749,60 @@ Proof.
...
+  - by specialize (c i).
+  - apply MC_non_initial_valid_consistent.
+    split; [| done].
+    by apply valid_trace_last_pstate in Hfvt.
+Qed.
+
+Lemma consistent_valid_state_prop :
+  forall (s : composite_state MCVLSM) (i : index),
+    valid_state_prop MC_composite_vlsm s -> st_rs (s i) <> None ->
+      consistent s.
```
palmskog, line 780:
```suggestion
    valid_state_prop MC_composite_vlsm s ->
    st_rs (s i) <> None ->
      consistent s.
```
I think this is much easier to read with one premise per line.

palmskog (approved):
Looks reasonable, but please address my comments.

theories/Examples/Tutorial/MuddyChildrenRounds.v
wkolowski: Changed.

theories/Examples/Tutorial/MuddyChildrenRounds.v
palmskog: Fine by me to print (only) these records using constructors, but I don't think it's essential.

wkolowski: Ok, I won't change it for now, but I'll keep it in mind.

PR #333, opened 2023-11-23
Turn some more asserts into lemmas in the Muddy Children example

wkolowski:
Besides the titular changes, there are also small improvements in a few unrelaed `assert`s.

### DIFF
--- a/theories/Examples/Tutorial/MuddyChildrenRounds.v
+++ b/theories/Examples/Tutorial/MuddyChildrenRounds.v
@@ -952,11 +952,10 @@ Proof.
   pose proof (Hs' := Hs).
   apply MC_composite_invariant_preservation in Hs.
   destruct (Hs (msg_index m)) as [Hinit | Hinvariant].
-  - assert (Hsminit : MC_initial_state_prop (s (msg_index m))) by apply Hinit.
-    unfold MC_initial_state_prop in Hsminit.
-    unfold MC_no_equivocation in Hnoequiv.
+  - unfold MC_no_equivocation in Hnoequiv.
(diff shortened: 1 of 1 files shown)


theories/Examples/Tutorial/MuddyChildrenRounds.v
```diff
@@ -1078,6 +1077,16 @@ Proof.
...
 
+Lemma st_obs_finite_valid_trace_from_to :
+  forall (s1 s2 : composite_state MCVLSM) (tr : list transition_item) (i : index),
+    finite_valid_trace_from_to MC_composite_vlsm s1 s2 tr ->
+      st_obs (s1 i) ≡ st_obs (s2 i).
+Proof.
+  intros.
+  apply MC_in_futures_preserves_obs_equiv.
+  by exists tr.
+Qed.
```
palmskog, line 1088:
```suggestion
Proof. by intros; apply MC_in_futures_preserves_obs_equiv; exists tr. Qed.
```
Seems like this is so short there is no need to use multiple lines.

palmskog (approved):
Fine by me, but I think we are reaching diminishing returns here.

theories/Examples/Tutorial/MuddyChildrenRounds.v
wkolowski: Changed.

wkolowski:
@palmskog Yes, I think there isn't much more that can be done with the Muddy Children example. I will open one more tiny PR to use constructors for records both in code and for printing and then switch to other tasks.

PR #336, opened 2023-11-24
add formulas example

palmskog:
Please review if headers and section names make sense.


wkolowski (approved):
Headers and sections for `Formulas.v` could be improved, as currently almost everything is subsumed under the header "VLSM for binary connective rules".

I opened a PR #337 with some suggestions (including some code moving).

PR #338, opened 2023-11-27
Refactor byzantine traces - comments

wkolowski:
Besides updating the comments, in this PR I changed the prefix `all_messages` (which occurred in names related to `emit_any_msg_vlsm`) to `emit_any_message_vlsm`.

### DIFF
--- a/theories/Core/ByzantineTraces.v
+++ b/theories/Core/ByzantineTraces.v
@@ -7,7 +7,8 @@ From VLSM.Core Require Import PreloadedVLSM.
 
   In this section, we introduce two definitions of Byzantine traces,
   prove that they are equivalent (lemma [byzantine_alt_byzantine_iff]),
-  and equivalent with traces on the corresponding pre-loaded VLSM
+  and then prove that both definitions are equivalent with the definition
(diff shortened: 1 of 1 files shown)


theories/Core/ByzantineTraces.v
```diff
@@ -6,19 +6,19 @@ From VLSM.Core Require Import PreloadedVLSM.
 (** * VLSM Byzantine Traces
 
   In this section, we introduce two definitions of Byzantine traces,
-  then show them equivalent (Lemma [byzantine_alt_byzantine_iff]),
+  prove that they are equivalent (lemma [byzantine_alt_byzantine_iff]),
   and equivalent with traces on the corresponding pre-loaded VLSM
```
traiansf, line 10: Maybe it didn't sound great before, but now it seems a bit fractured. Maybe something along the lines of:
```suggestion
  then show that both definitions are equivalent with the definition of a valid trace on the corresponding pre-loaded VLSM
```

traiansf (approved):

theories/Core/ByzantineTraces.v
wkolowski: Changed.

PR #341, opened 2023-11-27
Uniformize coqdoc module titles, add prefix

palmskog:
This is both to make coqdoc module titles consistent and make the coqdoc table of contents more easy to navigate.

### DIFF
--- a/theories/Core/Equivocation/TraceWiseEquivocation.v
+++ b/theories/Core/Equivocation/TraceWiseEquivocation.v
@@ -6,9 +6,9 @@ From VLSM.Core Require Import VLSM Composition ProjectionTraces.
 From VLSM.Core Require Import Equivocation.
 From VLSM.Lib Require Import Preamble StdppExtras.
 
-(** * VLSM Trace-wise Equivocation
+(** * Core: VLSM Trace-Wise Equivocation
 
-  In this section we define a more precise notion of message equivocation,
+  In this module, we define a more precise notion of message equivocation,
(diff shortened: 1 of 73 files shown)


traiansf (approved):
Looks reasonable to me, but wouldn't it look strange in the TOC to have so many of them prefixes by `Core`?

palmskog:
@traiansf actually, the TOC is fairly evenly split between "Utility" and "Core", with quite a lot of "Tutorial" at the end. I think it's at least an improvement visually from before.

theories/Core/Equivocation/TraceWiseEquivocation.v
```diff
@@ -6,9 +6,9 @@ From VLSM.Core Require Import VLSM Composition ProjectionTraces.
 From VLSM.Core Require Import Equivocation.
 From VLSM.Lib Require Import Preamble StdppExtras.
 
-(** * VLSM Trace-wise Equivocation
+(** * Core: VLSM Trace-Wise Equivocation
```
wkolowski, line 9: Maybe just spell it "tracewise", with one word?

wkolowski (commented):
Changes to comments, moving sections, and other changes of this kind are ok, but I really don't like the prefixes ("Core", "Utility", etc.) in top-level headers. They look rather repetitive. Maybe these prefixes could be added automatically during documentation generation?

palmskog:
@wkolowski we don't really have time to hack together documentation generation features in time for the next release. In the future, we are probably going to split files further at top level, so then it will become less "repetitive".

PR #343, opened 2023-11-28
Move plan-related code from casper-cbc-proofs

wkolowski:
In this PR I'm moving the contents from casper-cbc-proof's `Core/Plans.v` into this repo's `Core/Plans.v`, following some similar moving from PR #336.


palmskog:
I don't see any compelling reason to put this code here now, but I won't request any changes, I will let @traiansf decide.

traiansf (approved):
I am going to approve it since it's already done and it should bring no harm; however, @wkolowski please try to take @palmskog 's suggestion for future moves between the repositories.

PR #344, opened 2023-11-28
Rename `vtype` and `vmachine`

wkolowski:
In this PR I rename `vtype` to `vlsm_type` and `vmachine` to `vlsm_machine`. This is hopefully the last step in the process of removing the v-named definitions.

I think this PR should be included in the upcoming release because one of the reviewers was somewhat confused by names like `vTrace`. And if we indeed want this change, it would be nice to make it before updating the code listings in papers and paper maps.

The commands used for renaming:
```bash
sed -i 's/vtype/vlsm_type/g' $(find -name "*.v")
sed -i 's/vmachine/vlsm_machine/g' $(find -name "*.v")
```
@palmskog @bmmoore Any objections?


traiansf (commented):
I personally have no objection, but I think Karl should approve this.

palmskog (changes requested):
@wkolowski it doesn't really matter to me much about vmachine/machine, but `type` is a terrible keyword to use in a Coq library. New users will have no idea about the remarkable differences between `type` and `Type`, and even experienced users may get confused.

If `vtype` is not satisfactory, it would have to be something longer than `vlsm_type`.

palmskog:
In isolation, it doesn't much matter to me whether it says `machine` or `vmachine`, but I don't see any consistency in this PR if only `vmachine` but not `vtype` is changed. At least with the `v` prefix, there was something to hold together the record projections. As hinted at in the original review, `vlsm_machine` and `vlsm_type` would at least be consistent and a clear improvement on the current situation.

wkolowski:
@palmskog I changed it to `vlsm_type` and `vlsm_machine` as discussed during the meeting.

palmskog (approved):
`vlsm_type`/`vlsm_machine` is fine by me.

bmmoore (changes requested):
I agree with @palmskog, an unprefixed `type` is too confusing. Besides the confusion with `Type` and the general idea of Coq types, I think the single word "type" isn't a great fit for getting a package of several types (the VLSMType), and using a well-established term like "type" without any prefix or suffix suggests too strongly that it is supposed to be a great fit.

I think "vlsm_type" is a fine name, and understood "vtype" as running that name through the convention of compressing "vlsm_" to "v" to save characters, but I suppose that convention was mostly demonstrated - and the space savings mostly productive - through names like vstate and vlabel and vtransition that we have already changed.

So, `machine` is fine by me but not `type`. Leaving `vtype` and `vmachine` is also fine, but if you prefer expanding to `vlsm_type` and `vlsm_machine` that works too.

PR #347, opened 2023-12-04
Tweak printing records in ELMO

wkolowski:
The command `#[export] Unset Printing Records.` used to force printing of some record types using constructors in `BaseELMO.v` was a bit of an overkill, because it was exported out of the module and then reexported from `AllELMO.v`, so it could possibly wreak some havoc for the users.

In this PR I changed it to the more fine-grained command `Add Printing Constructor`. This makes `State`, `Observation` and `Message` be displayed using the constructor, and this setting is exported to the outside world, but all other records are unaffected.

### DIFF
--- a/theories/Examples/ELMO/BaseELMO.v
+++ b/theories/Examples/ELMO/BaseELMO.v
@@ -46,7 +46,7 @@ with Message : Type := MkMessage
   types, using their constructor names, instead of as records with the
   {| ... |} notation.
 *)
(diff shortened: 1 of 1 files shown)


bmmoore (approved):
Much better to do it this way.

It is unfortunate and a bit inconsistent that `Add` couldn't take the `#[export]` attribute, but it seems repeating it after closing the section is the only option.

PR #350, opened 2023-12-05
Fix the axiom diagnostics script

wkolowski:
The axiom diagnostics script is got broken at some point in the past. I'm suspecting it's because of the directory structure changes, but it could also be because of my setup.

Since the script should work properly for users running `make axioms`, in this PR I'm offering a fix. Please check locally if it works for you.


bmmoore (changes requested):
On Debian, /bin/sh is not bash but a simple posix shell, so the `[[` on line 26 is not allowed, and the `==` on lines 26 and 106 don't do the right thing.
I think just testing `[ "$3" = true ]` on line 26 and `[ $keep_tmp = false ]` on 106 works.

On windows coq platform uses cygwin to run make, but coqc can't handle files under a cygwin path "`/tmp`". This can be fixed by making the temporary directory locally, with `tmp=$(mktemp -d -p.)`

wkolowski:
@bmmoore I applied your suggestion regarding the temporary directory, but the other ones didn't work for me. Instead I decided to switch it from using `sh` to using `bash`. Please test if it works on your machine and re-review.

bmmoore (approved):
Looks good

PR #353, opened 2023-12-08
Fix 8.19

palmskog:
These changes are entirely based on what is required to be compatible with 8.19 while maintaining backwards compatibility. For example, the latest stdpp removed the `wf` shorthand.

### DIFF
--- a/coq-vlsm.opam
+++ b/coq-vlsm.opam
@@ -17,7 +17,7 @@ build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
   "dune" {>= "3.5"}
   "coq" {>= "8.16"}
-  "coq-stdpp" {= "1.9.0"}
+  "coq-stdpp" {>= "1.9.0"}
   "coq-itauto"
   "coq-equations"
 ]
--- a/meta.yml
+++ b/meta.yml
@@ -52,9 +52,9 @@ supported_coq_versions:
 dependencies:
 - opam:
     name: coq-stdpp
-    version: '{= "1.9.0"}'
+    version: '{>= "1.9.0"}'
   description: |-
-    [Coq-std++](https://gitlab.mpi-sws.org/iris/stdpp/) 1.9.0
+    [Coq-std++](https://gitlab.mpi-sws.org/iris/stdpp/) 1.9.0 and later
 - opam:
     name: coq-itauto
   description: |-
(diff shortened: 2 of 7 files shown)


coq-vlsm.opam
```diff
@@ -17,7 +17,7 @@ build: ["dune" "build" "-p" name "-j" jobs]
 depends: [
   "dune" {>= "3.5"}
   "coq" {>= "8.16"}
-  "coq-stdpp" {= "1.9.0"}
+  "coq-stdpp" {>= "1.9.0"}
```
wkolowski, line 20: Again, intended?

meta.yml
```diff
@@ -52,9 +52,9 @@ supported_coq_versions:
 dependencies:
 - opam:
     name: coq-stdpp
-    version: '{= "1.9.0"}'
+    version: '{>= "1.9.0"}'
   description: |-
-    [Coq-std++](https://gitlab.mpi-sws.org/iris/stdpp/) 1.9.0
+    [Coq-std++](https://gitlab.mpi-sws.org/iris/stdpp/) 1.9.0 and later
```
wkolowski, line 57: Again.

theories/Examples/ELMO/ELMO.v
```diff
@@ -429,9 +429,9 @@ Lemma ELMO_input_constrained_transition_inj :
 Proof.
   intros l s om s' om' [(_ & _ & Hvalid) Ht] l0 s0 om0 om'0 [(_ & _ & Hvalid0) Ht0].
   by inversion Hvalid; subst; cbn in Ht;
-    injection Ht as [= <- <-];
-    inversion Hvalid0; subst; inversion Ht0;
-    replace s0 with s by (apply eq_State; done).
+  injection Ht as [= <- <-];
+  inversion Hvalid0; subst; inversion Ht0;
+  [replace s0 with s by (apply eq_State; done)|].
```
wkolowski, line 434: Note that all of this is one big chain that follows the `by inversion` from the line above, so I think it should be indented by 2 spaces, according to our current coding conventions.

meta.yml
palmskog: Yes, all of these are intended since Stdpp 1.9.0 doesn't work on Coq 8.19. So to allow compatibility with Coq 8.19 and beyond, I have to relax the restriction to `coq-stdpp.1.9.0` only.

PR #354, opened 2023-12-08
Bits and pieces...

traiansf:
This is what I could salvage from my recent attempt on bridging limited equivocation results. I would like this to be approved by @palmskog before merging it, as it can definitely be postponed until after the release.

### DIFF
--- a/theories/Core/MessageDependencies.v
+++ b/theories/Core/MessageDependencies.v
@@ -2,7 +2,7 @@ From VLSM.Lib Require Import Itauto.
 From stdpp Require Import prelude.
 From VLSM.Lib Require Import Preamble ListExtras.
 From VLSM.Core Require Import VLSM VLSMProjections Composition ProjectionTraces.
-From VLSM.Core Require Import SubProjectionTraces Equivocation EquivocationProjections. 
+From VLSM.Core Require Import SubProjectionTraces Equivocation EquivocationProjections.
 
 (** * Core: VLSM Message Dependencies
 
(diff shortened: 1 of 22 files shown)


palmskog (approved):
I think this PR has some good refactorings and renamings and overall would be a net benefit to a release. In particular, the `ELMO_A` names were awkward and the `finset` stuff placement is much better.

theories/Core/MessageDependencies.v
```diff
@@ -1261,4 +1261,50 @@ Proof.
...
 Qed.
 
+(** We also define the case in which a transition doesn't forget equivocation. *)
+Definition transition_preserves_global_equivocation
+  (s : composite_state IM) (item : composite_transition_item IM) : Prop :=
+  forall (v : validator),
+    msg_dep_is_globally_equivocating IM message_dependencies sender s v ->
+    msg_dep_is_globally_equivocating IM message_dependencies sender (destination item) v.
+
+Inductive TraceMonotonicGlobalEquivocation :
```
wkolowski, line 1271: I think "monotone" is more idiomatic in English. This also applies to the names below.
```suggestion
Inductive TraceMonotoneGlobalEquivocation :
```

theories/Examples/ELMO/ELMO.v
```diff
@@ -21,16 +22,46 @@ Section sec_ELMO.
...
-  `{FinSet Address Ca}
   (threshold : R)
-  `{!ReachableThreshold Address Ca threshold}
   `{finite.Finite index}
   (idx : index -> Address)
-  `{!Inj (=) (=) idx}.
+  `{!Inj (=) (=) idx}
+  `{!ReachableThreshold (Message_validator idx) (listset (Message_validator idx)) threshold}.
+
+#[local] Instance Address_reachable_threshold :
```
wkolowski, line 36: I'd suggest to call this `ReachableThreshold_Address`. Also, why `#[local]`?

```suggestion
#[local] Instance ReachableThreshold_Address :
```

wkolowski (approved):
Looks ok. I opened a PR with suggestions: #357

wkolowski:
@traiansf Also, before merging make sure to prepare fixes for casper-cbc-proofs, as some things break there.

PR #356, opened 2023-12-08
avoid collision between names in Voting.v and Paxos.v

palmskog:
This is to allow people to `Import` both `Voting` and `Paxos` and allow names to be unique. I also uniformized some VLSM-related names and annotations.


palmskog:
@bmmoore are you OK with merging this?

PR #372, opened 2023-12-13
Cleanup the use of `pre_loaded_with_all_messages_vlsm`

wkolowski:
In this PR I:
- replace the remainig uses of valid/preloaded that were missed in previous PRs
- remove a lot of abbreviations for `pre_loaded_with_all_messages_vlsm` which are not longer needed

### DIFF
--- a/theories/Core/Equivocation.v
+++ b/theories/Core/Equivocation.v
@@ -1029,7 +1029,6 @@ Proof.
   intros l s im s' om Htrans msg.
   rename Htrans into Htrans'.
   pose proof Htrans' as [[Hproto_s [Hproto_m Hvalid]] Htrans].
-  set (preloaded := pre_loaded_with_all_messages_vlsm vlsm) in * |- *.
   pose proof (valid_state_has_trace _ _ Hproto_s)
     as [is [tr [Htr Hinit]]].
   pose proof (Htr' := extend_right_finite_trace_from_to _ Htr Htrans').
@@ -2615,7 +2614,6 @@ Context
   {message : Type}
   `{EqDecision message}
   (X : VLSM message)
-  (PreX := pre_loaded_with_all_messages_vlsm X)
   `{HasBeenSentCapability message X}
   `{HasBeenReceivedCapability message X}
   .
@@ -2625,7 +2623,7 @@ Definition state_received_not_sent (s : state X) (m : message) : Prop :=
 
 Lemma state_received_not_sent_trace_iff
   (m : message)
-  (s is : state (PreX))
+  (s is : state (pre_loaded_with_all_messages_vlsm X))
   (tr : list transition_item)
   (Htr : finite_constrained_trace_init_to X is s tr)
   : state_received_not_sent s m <-> trace_received_not_sent_before_or_after tr m.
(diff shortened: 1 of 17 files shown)


theories/Core/Equivocation.v
```diff
@@ -2625,7 +2623,7 @@ Definition state_received_not_sent (s : state X) (m : message) : Prop :=
 
 Lemma state_received_not_sent_trace_iff
   (m : message)
-  (s is : state (PreX))
+  (s is : state (pre_loaded_with_all_messages_vlsm X))
```
traiansf, line 2626: Here, and elsewhere: why not state X

theories/Core/Equivocation.v
```diff
@@ -2673,8 +2671,8 @@ Definition cannot_resend_message_stepwise_prop : Prop :=
 
 Lemma cannot_resend_received_message_in_future
   (Hno_resend : cannot_resend_message_stepwise_prop)
-  (s1 s2 : state (PreX))
-  (Hfuture : in_futures PreX s1 s2)
+  (s1 s2 : state (pre_loaded_with_all_messages_vlsm X))
+  (Hfuture : in_futures (pre_loaded_with_all_messages_vlsm X) s1 s2)
```
traiansf, line 2675: maybe have an `in_futures_constrained` definition?

theories/Core/Equivocation/TraceWiseEquivocation.v
```diff
@@ -96,7 +95,7 @@ Lemma elem_of_equivocating_senders_in_trace
   : v ∈ equivocating_senders_in_trace tr <->
     exists (m : message),
     (sender m = Some v) /\
-    equivocation_in_trace PreFree m tr.
+    equivocation_in_trace (pre_loaded_with_all_messages_vlsm (free_composite_vlsm IM)) m tr.
```
traiansf, line 98: Here I actually think it should work with `Free`.

theories/Core/EquivocationProjections.v
```diff
@@ -434,8 +433,9 @@ Context
   by the free composition (pre-loaded with all messages), then it can also be
   emitted by the component corresponding to its sender (pre-loaded with all messages).
 *)
-Lemma can_emit_projection
-  : can_emit PreFree m -> can_emit (pre_loaded_with_all_messages_vlsm (IM j)) m.
+Lemma can_emit_projection :
+  can_emit (pre_loaded_with_all_messages_vlsm (free_composite_vlsm IM)) m ->
```
traiansf, line 437: maybe introduce `constrained_message_prop` and use it here?

traiansf (approved):
Reasonable, but please address my comments.

theories/Core/Equivocation.v
wkolowski: Even though `state X` and `state (pre_loaded_with_all_messages_vlsm X)` are computationally equal, making this change sometimes breaks proofs, so I didn't do it. I'll leave it for the future.

theories/Core/Equivocation.v
wkolowski: I'll leave it as a TODO for the future.

theories/Core/Equivocation/TraceWiseEquivocation.v
wkolowski: Indeed. Thanks!

theories/Core/EquivocationProjections.v
wkolowski: It is already defined, but I didn't include it in these refactorings, because it's a little confusing (since it's not just `valid_message_prop (pre_loaded_with_all_messages_vlsm X)`.

PR #379, opened 2023-12-14
An attempt at support for Nix shell/build

wkolowski:


wkolowski:
This PR didn't make it into the release, so I'm closing for now. It can be reopened in case Nix support is ever needed.

PR #384, opened 2023-12-20
Farewell refactorings

wkolowski:
It's been an honour to be able to work on this project, and especially to do all the refactorings :) This project is where I learned the "industrial" side of proof engineering.

In this PR I refactor four proofs, one of whom had wrong bullets at the top level, and the three remaining ones I found offensive for some reason long forgotten :)

### DIFF
--- a/theories/Core/Equivocators/SimulatingFree.v
+++ b/theories/Core/Equivocators/SimulatingFree.v
@@ -110,53 +110,46 @@ Lemma generalized_equivocators_finite_valid_trace_init_to_rev
     finite_valid_trace_init_to CE is s tr /\
     finite_trace_last_output trX = finite_trace_last_output tr.
 Proof.
-  assert (HinclE : VLSM_incl CE (preloaded_with_all_messages_vlsm FreeE))
-    by apply constrained_preloaded_vlsm_incl_preloaded_with_all_messages.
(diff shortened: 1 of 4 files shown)


theories/Core/Equivocators/SimulatingFree.v
```diff
@@ -169,54 +162,41 @@ Proof.
     cbn in Hesom', Hes_pr_eqv.
     rewrite Hes_pr_eqv in Hesom'.
     cbn in Ht.
-    destruct (transition _ _ _) as (si', _om) eqn: Hteqv.
-    inversion Ht. subst sf _om. clear Ht.
-    inversion Hesom'. subst es' om'. clear Hesom'.
-    match type of Happ_extend with
-    | ?H -> _ => cut H
+    destruct (transition _ _ _) as [si' _om] eqn: Hteqv.
```
traiansf, line 165: I think `destruct transition` also works directly

traiansf (approved):

theories/Core/Equivocators/SimulatingFree.v
wkolowski: Maybe it works, but it shouldn't and I consider it a bad practice. After all, unapplied `transition` is a function, but you should only (be able to) `destruct` inductives/records.

